PR #11, opened 2015-10-07
OptiMimi optimization wrapper

jrising:
Here's the current, working version the optimization wrapper.  It will probably still change as I expand the applications beyond my dummy example, but I wanted to share it with you before its in stone.  Some notes:

1. There's a documentation file, but I haven't added it into the wiki.  It's in docs/optimimi.md.
2. I added handling for models without timesteps, which define a 'simulate' function rather than a 'timestep' function.  This is partly for my dummy example, but I think I had another reason at some point.
3. My changes shouldn't have any impact on existing Mimi models, since they're wrapped up in a separate module that you can include.

My next step is to get automatic differentiation working.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,5 @@
-# Mimi.jl v0.1.1 Release Notes
+# Mimi.jl v0.0.2 Release Notes
 * Make julia 0.4 compatible
 
-# Mimi.jl v0.1.0 Release Notes
+# Mimi.jl v0.0.1 Release Notes
 * Initial tagged version


davidanthoff:
Very cool! I haven't looked at it in detail, but I'll try to do so soon.

Couple minor things:
- CI fails right now because NLopt is not in REQUIRED, that should be easy to fix
- To add something to the documentation, just add a line to mkdocs.yml

jrising:
I updated those two.  It's a shame we need to REQUIRE NLopt, when this is supposed to be a optional piece.  But the changes are going to need to go deeper with ForwardDiff.jl, so I don't want to make this a separate package.

jrising:
Okay, these new commits now implement automatic differentiation (using ForwardDiff) in Mimi/OptiMimi.  You have to both use the `autodiffable` optional argument when creating a Model, and define all of the components with @defcompo (rather than @defcomp) to get that functionality, but if you don't, everything still works, including optimization which falls back on derivative-free algorithms.

Also, I replaced the 'String's with 'AbstractString's in Mimi, to clear up the warnings in v0.4.

jrising:
And now with v0.3 compatibility (including the old API for ForwardDiff) and some unit tests to boot.  Enjoy!

davidanthoff:
Now had a look, really cool!

I agree, it would actually be nicer, if we didn't have to take a dependency on the optimization packages. I think generally it would be cool if things were modular enough that this stuff could live in its own package. We could then also have packages like MCMC fitting packages that fit a Mimi model etc.

I also had a go at supporting ForwardDiff in the core package, but using parametric types. My work in progress is in #13. The approach is modeled on what you did, but a bit more general. In particular, it should work for any alternative numeric type, doesn't require any new dependency in the Mimi package and it doesn't require an alternative ``@defcompo`` macro, but should work with any Mimi component. Do you think that kind of approach would work with your optimization stuff?

I actually think the ability to use other alternative number types like for example [interval arithmetic](https://github.com/dpsanders/ValidatedNumerics.jl) types could be hugely beneficial above and beyond the optimization stuff. Another application I can think of is a special number type that throws an error whenever the result of a computation would be NaN. This could be used for debugging: some of the most annoying bugs I've dealt with occurred when my final result was a NaN, and then I tried to figure out where that problem started. Running such a model with a number type that throws an error the first time a NaN occurs would be super useful.

Oh, and should we just ditch julia 0.3 support? Will just be a pain to support both platforms, and now that 0.4 is released, I'd be all for just making the minimum julia version 0.4.

Maybe in terms of next steps, can you take a look at my PR and try to figure out whether that would work? There are still a bunch of things I need to do before I could merge that, and one is to get an OK from you.

jrising:
I think that a parametric type would probably work well.  I think we might still need to require that it's a subclass of Number, at least when using DualNumbers, but even that could probably be relaxed if the developer wanted something else in the arrays.

I don't think we need to drop 0.3 support quite yet.  A lot of people still have Julia 0.3 on their machines, and we can wait until there's a real incompatibility.

That all said, I think you should accept my pull request and make your modifications to that.  I updated the code to be more compatible with 0.4, and I would have to do my merge from scratch.  Plus, you can use the OptiMimi tests to make sure that DualNumbers are working under your new system.

davidanthoff:
I just realized that you did all your changes on master, but I had to push some small things on master in the meantime and tag a new version for some other people that are using Mimi. So we will have to sort out that branch situation...

In general, PRs should never come as commits to master, the work for a PR should always be in a separate/new branch. Much easier that way, the whole workflow is described [here](http://docs.julialang.org/en/release-0.4/manual/packages/#code-changes), and then things like other updates to master while we discuss this PR are super easy to handle.

To fix this, we will need to move your commits to a new branch, change this PR over to the new branch, and then you need to delete your old master branch and sync up a new master branch with the master branch in this repository. If this sounds like gibberish, I can sort out the branch situation once I have a minute. The whole procedure is not super trivial, unless you know git fairly well...

I'll get back to the substantive stuff in this PR once I have more time!

jrising:
It's no problem.  I've rebased to your changes and moved to a different
branch and initiated a new pull request.


 - James

PR #13, opened 2015-10-20
WIP: Parametric number type

davidanthoff:
This is a work in progress, not yet ready to be merged.

This makes all the generated types from the ``@defcomp`` macro parametric types, so that one can specify a type that should be used for any parameter or variable that is defined as a Number. The idea is that one can easily choose between say a ``Float64``, some kind of dual number or something entirely different as the type for numbers.

Things that still need to be done:
- [x] Compare performance to the previous implementation. I don't think this should be any slower, but we should make sure that is so.
- [x] Expose the whole functionality via an alternative constructor for a ``Model``.
- [ ] Write tests.
- [ ] Write documentation.
- [ ] Update NEWS.
- [ ] Make sure this actually is flexible enough to e.g. enable the use of ForwardDiff.

There are also at least two design questions that are not trivial:
- The current design will require a choice about the type to be used at the time of construction of the model object. In some ways that is not ideal, it would be better if one could create the structure of the model without making that choice, and then could specify the type when one runs the model. This would require some kind of object model for the model structure, and then a separate step where the objects that hold the state of the model are constructed just before the model is run. That last step could then decide what type to use for numbers. I think overall that kind of design makes a lot of sense (might also make the whole parameter handling thing easier), but for now I'm probably going to merge this before that is resolved and tackle this issue later.
- The current design will require that all parameters have the same type for numbers as the main model. This is not ideal. If the model for example uses some kind of dual number, it would be better to be able to still pass exogenous parameters as a simple ``Array{Float64}``. Solving this is really tricky if one still wants performance, so also for a later time and another PR.

### DIFF
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -500,8 +500,8 @@ macro defcomp(name, ex)
 
 		type $(esc(symbol(string(name, "Impl")))){T} <: $(esc(symbol(name)))
 			nsteps::Int
-			Parameters::$(esc(symbol(string(name,"Parameters"))))
-			Variables::$(esc(symbol(string(name,"Variables"))))
+			Parameters::$(esc(symbol(string(name,"Parameters")))){T}
+			Variables::$(esc(symbol(string(name,"Variables")))){T}
 			Dimensions::$(esc(symbol(string(name,"Dimensions"))))
 
 			function $(esc(symbol(string(name, "Impl")))){T}(::Type{T}, indices)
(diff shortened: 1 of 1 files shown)


davidanthoff:
@jrising it looks as if this actually doesn't work on julia 0.3 because they changed how constructors work and I'm relying on the new behavior. I'll try to see whether I can figure out an easy way around it, but otherwise I'll just merge #14 (assuming you are on julia 0.4, right? I know the other current users of Mimi, and that would work for them as well).

jrising:
Yup, I'm on 0.4.  I guess it is time to drop 0.3!

PR #15, opened 2015-10-22
OptiMimi optimization wrapper

jrising:
Here's OptiMimi as a separate branch, rebased to the changes you've made in the meantime.

This is a follow up of the pull request here: https://github.com/davidanthoff/Mimi.jl/pull/11


davidanthoff:
Excellent.

I now combined your work with the parametric type stuff from #13, you can see the result in #16. With that it should be easy to move the optimization stuff into its own package.

I think right now we should move the OptiMimi into its own package, mainly because of my time constraints. I need to get the redesign of the parameter system done as soon as possible because there are a whole lot of things (with deadlines) waiting for that, and that will be less work if I don't have to keep the optimization code in sync. I normally wouldn't mind, but right now I just don't have the time. With a separate OptiMimi package we can actually decouple things: OptiMimi can just adjust to changes in Mimi on its own schedule, and at the same time we can have a Mimi package with a master branch that is "ready to ship" at any time. Plus, at this moment you don't have to wait for me to review PRs around optimization (which would be slow), but can just press ahead in a OptiMimi repo.

jrising:
I'm fine with having OptiMimi as its own package.  I can pull it out, once we resolve these various pull requests (or you can do it if you want).

davidanthoff:
Alright, all my PRs are merged, so I think everything should be ready to move OptiMimi into a new package. Make sure to submit a PR to the README here to link to it once the repo is up!

jrising:
I've submitted OptiMimi as an official package (because, why not?), and
I'll update the README when it's accepted.


 - James

PR #16, opened 2015-10-23
WIP: optimization based on parametric types

davidanthoff:
This branch changes @jrising's OptiMimi module to use the parametric type work in #13. The main benefit is that the whole optimization work can be moved into its own package because there are no changes in Mimi required (beyond the parametric types) to get the optimization and the ForwardDiff stuff to work.

Various random comments:
- I removed the ``simulate`` function, one can just use a one timestep model instead. I don't want to expose too many different options to users at this point, plus there are a whole bunch of questions how this would compose (i.e. what happens if one creates a model with static and dynamic components?). We can revisit later once I have more time.
- I moved the conversion from ``Array{Float64]`` to ``Array{Number}`` that needs to happen for parameters out of the ``connectparameter`` function and into the test code. Two reasons: 1) the semantics of array parameters in Mimi are that they are by ref, i.e. you can set an array as a parameter, and then if you change a value in the array it should be picked up by the component. When the conversion occurs in ``connectparameter`` this doesn't work because a new copy is created. 2) I need to change the whole parameter setup in any case because it also doesn't work for some of the monte carlo stuff we need (and need ASAP...). I'll try to just resolve this in one go.
- In general, using ``Number`` as the type of fields for the ForwardDiff case is almost certainly not ideal. ``Number`` is an abstract type and what you end up with is lots of ``Array{Number}`` and instance members of type ``Number``. Both of these are really bad for [performance](http://docs.julialang.org/en/release-0.4/manual/performance-tips/#avoid-containers-with-abstract-type-parameters), mainly because everything gets boxed, and all calls end up using virtual calls and nothing can get inlined. I think a much better approach would be to create a model that uses the exact type that the ForwardDiff package uses. A couple of months ago that would have been a dual number, then inbetween a ``NDual``, but now they seem to have change it again, so I'm no longer sure.

In terms of next steps, I would suggest this: if @jrising thinks this design can work, I will merge #13 into master. Then @jrising can create a new package OptiMimi and move the code from his/this PR into that package. How does that sound?


jrising:
Great!  This is a really nice improvement.

I don't like that the aesthetics of setting up a model for automatic differentiation are messy (that you need to pass Number in to the Model constructor and create the parameters with Number arrays), but my solution wasn't any less messy.  Do we need people to be able to change parameters by changing the values in the array?  I think we can reduce the amount that users need to think about if we say "Any time you want to change the parameters of a model, you must call setparameter again."

If you don't want to do that, I might wrap something around the Mimi model in OptiMimi to do it for users.

I'll play around with using the ForwardDiff parameter type instead of Number.  This is another case where it probably makes sense for OptiMimi to provide a simplified interface.

Anyway, this looks great and I think you should merge away!

davidanthoff:
> I don't like that the aesthetics of setting up a model for automatic differentiation are messy (that you need to pass Number in to the Model constructor and create the parameters with Number arrays), but my solution wasn't any less messy.

Yep, I agree. I think the right moment to make the choice of number type is when one calls ``run``, so that one can construct a model, and then run the same model with different number types. I might try to implement that in one go with the new parameter stuff,  essentially what this requires is that calls to things like ``addcomponent`` and ``connectparameter`` etc. don't operate on the objects that will hold variables at runtime, but build up a blueprint of how the model should look like, and then the objects that hold the variables and parameters get constructed before one calls run. Like a compile step...

> Do we need people to be able to change parameters by changing the values in the array? I think we can reduce the amount that users need to think about if we say "Any time you want to change the parameters of a model, you must call setparameter again."

Yes, that might make more sense, I just need to think through it before we change it, I have a vague memory that I might have relied on the old behavior somewhere...

davidanthoff:
@jrising: I'll leave the branch ``da-optim`` around until you have moved its content to a new repo.

PR #20, opened 2015-12-06
Add component references system

jrising:
Here is the system for component references-- so that we can write, e.g.,
connectparameter(model, :MyComponent, :myparameter, :YourComponent, :yourparameter)
as
mycomponent[:myparameter] = yourcomponent[:yourparameter]

### DIFF
--- a/test/test_references.jl
+++ b/test/test_references.jl
@@ -6,7 +6,7 @@ using Mimi
 	intermed = Variable(index=[time])
 end
 
-function timestep(c::Foo, tt::Int)
+function timestep(c::Foo, tt)
     c.Variables.intermed[tt] = c.Parameters.input
 end
 
@@ -15,7 +15,7 @@ end
 	output = Variable(index=[time])
 end
 
-function timestep(c::Bar, tt::Int)
+function timestep(c::Bar, tt)
     c.Variables.output[tt] = c.Parameters.intermed[tt]
 end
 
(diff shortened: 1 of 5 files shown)


jrising:
Have you seen this before?  Only 32 bit windows failing, for a very simple model?  I thought it had to do with my use of 'pi' as a test number, but no.

test/test_references.jl
```diff
@@ -0,0 +1,33 @@
+using Base.Test
+using Mimi
+
+@defcomp Foo begin
+	input = Parameter()
+	intermed = Variable(index=[time])
+end
+
+function timestep(c::Foo, tt::Int)
```
davidanthoff: Try to remove the type annotation on ``tt`` here, i.e. change this line to ``function timestep(c::Foo, tt)``. I think what is happening is that ``Int`` defaults to ``Int64`` on x64 platforms, but to ``Int32`` on x86 systems. But Mimi calls ``timestep`` always with a ``Int64`` for the second parameter. So when you declare with ``Int``, multi dispatch will fall back to the generic timestep on x86 systems, and your timestep function is never called.

test/test_references.jl
```diff
@@ -0,0 +1,33 @@
...
+    c.Variables.intermed[tt] = c.Parameters.input
+end
+
+@defcomp Bar begin
+	intermed = Parameter(index=[time])
+	output = Variable(index=[time])
+end
+
+function timestep(c::Bar, tt::Int)
```
davidanthoff: Same here.

davidanthoff:
Great, this is really cool!

PR #24, opened 2016-03-27
Move type creation from macro to normal runtime phase

davidanthoff:
@jrising and @FrankErrickson: This changes some fundamental things under the hood, but should not break any existing code that uses Mimi. I tried it with fund.jl, mimi-sneasy.jl, mimi-dice-2013.jl and mimi-rice-2010.jl and those all seem to work. I believe you both have code basis that use Mimi that I don't have access to, could you try to run those with the version of Mimi in this PR and let me know if there are any problems?

The major benefit for now is that one can redefine a component with this version without exiting the julia process. More medium/longterm this PR lays the foundation for a whole bunch of other improvements. Plus the code should be a fair bit easier to maintain because now less stuff is happening in the macro itself, which always was a pain to develop/debug.

Thanks!

### DIFF
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -382,13 +382,6 @@ end
 Define a new component.
 """
 macro defcomp(name, ex)
-    dimdef = Expr(:block)
-    dimconstructor = Expr(:block)
-
-    pardef = Expr(:block)
-
-    vardef = Expr(:block)
-    varalloc = Expr(:block)
     resetvarsdef = Expr(:block)
 
     metavardef = Expr(:block)
@@ -398,9 +391,6 @@ macro defcomp(name, ex)
     for line in ex.args
         if line.head==:(=) && line.args[2].head==:call && line.args[2].args[1]==:Index
             dimensionName = line.args[1]
-
-            push!(dimdef.args,:($(esc(dimensionName))::$(esc(UnitRange{Int64}))))
-            push!(dimconstructor.args,:(s.$(dimensionName) = UnitRange{Int64}(1,indices[$(QuoteNode(dimensionName))])))
 
             push!(metadimdef.args, :(metainfo.adddimension(module_name(current_module()), $(Expr(:quote,name)), $(QuoteNode(dimensionName)) )))
         elseif line.head==:(=) && line.args[2].head==:call && line.args[2].args[1]==:Parameter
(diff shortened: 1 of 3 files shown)


jrising:
Wonderful!  Sorry for the delay-- I just checked it against one of my models, and it seems solid.

davidanthoff:
Excellent, thanks!

jrising:
Found a difference!  This is certainly the result of my hacks beyond the "official" Mimi API, but it points to the need for new official functions.

Sometimes I need to get the dimensions of a variable within a model.  To get the dimensions of a *parameter*, I can ask `size(model.parameters[name].values)` (which isn't great either, because it wouldn't work if there were any parameter name conflicts.  But for a variable, the only way I was able to figure out how to do it was to call `meta[eval(component)].variables[name].dimensions` (where I need to evaluate component because I want to otherwise pass component around as a symbol.

Now that's changed, so I need to say `meta[(:Main, component)]...`.  What is `:Main`, and when would that change?

I'm not actually sure that it's an improvement to have the components indexed by symbol and not by object.  In the past, I've had need to create a "PassThrough" component, which just copies its parameter to its variable, so that I can take a single parameter and connect it into many components.  But it would be nice to be able to initialize more than one PassThrough, with feeding different parameters into each one.  I guess I should just use the addcomponent(m, t, name) syntax for this.

None of this is critical now.  But I think you should produce an request-for-comments-style document which offers a system to:
 - Get index counts and values for dimensions, by name
 - Get index definitions for parameters and variables, by component and name

Thanks!

davidanthoff:
The ``meta`` dict is now indexed by a tuple of module and component name. So if you do a ``@defcomp`` inside a julia module, then Mimi now also stores that component as in that module. Main just is the default module.

>  In the past, I've had need to create a "PassThrough" component, which just copies its parameter to its variable, so that I can take a single parameter and connect it into many components.

I.e. you want to connect an external parameter to multiple components? I *think* you can get there by not using ``setparameter`` at all, but instead first adding the external parameter via ``addparameter`` (at that point no component is picking it up), and then connecting as many components to that external parameter as you want via ``connectparameter``. But that whole design is not good right now, definitely something we should iterate on.

> Get index counts and values for dimensions, by name
> Get index definitions for parameters and variables, by component and name

Completely agreed! There is some tricky issue here that some of this information is tied to the component definition, and some of it is only present in a concrete model. I'm also not sure I made the right trade of on that front... Do you want to open issues for these things, so that we don't forget them?

jrising:
I'll try the `addparameter`-`connectparameter` combination.  That seems sensible.  Maybe something to add a FAQ on?

I'll add issues for the index query items.

PR #29, opened 2016-04-26
add unit handling and connection checking

jrising:
This is the component-to-component part of the unit checking system.  You add new units like this:

```waterdemand = Parameter(index=[regions, time], units="1000 m^3")```

and when you call `connectparameter` with two components, it checks that the units match, allowing a special `any` unit that matches anything, but otherwise requiring that the units are the same or both missing.  I have all my calls like this through lines like:

```allocation[:waterdemand] = waterdemand[:totaldemand]```

This does not (1) handle any unit checking anywhere except for through `connectparameter` (and not the version of `connectparameter` that takes an already-added parameter as an argument, (2) have any logic for the `UnitedObject` that I mentioned before, nor (3) apply the unit logic to equations.

### DIFF
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -275,13 +275,16 @@ function addparameter(m::Model, name::Symbol, value)
     end
 end
 
-function connectparameter(m::Model, target_component::Symbol, target_name::Symbol, source_component::Symbol, source_name::Symbol)
+function connectparameter(m::Model, target_component::Symbol, target_name::Symbol, source_component::Symbol, source_name::Symbol; ignoreunits::Bool=false)
     c_target = m.components[target_component]
     c_source = m.components[source_component]
 
     # Check the units, if provided
-    @assert unitcheck(getmetainfo(m, target_component).parameters[target_name].unit,
-                      getmetainfo(m, source_component).variables[source_name].unit) "Units of $source_component.$source_name do not match $target_component.$target_name."
+    if !ignoreunits &&
+        !unitcheck(getmetainfo(m, target_component).parameters[target_name].unit,
+                   getmetainfo(m, source_component).variables[source_name].unit)
+        throw(ErrorException("Units of $source_component.$source_name do not match $target_component.$target_name."))
+    end
 
     setfield!(c_target.Parameters, target_name, getfield(c_source.Variables, source_name))
     push!(m.parameters_that_are_set, string(target_component) * string(target_name))
@@ -291,9 +294,9 @@ end
 """
 Default string, string unit check function
 """
-function unitcheck(one::Union{UTF8String, ASCIIString}, two::Union{UTF8String, ASCIIString})
-    # True if both "", either are "any", or they match
-    return one == "any" || two == "any" || one == two
+function unitcheck(one::AbstractString, two::AbstractString)
+    # True if and only if they match
+    return one == two
 end
 
 """
@@ -440,9 +443,9 @@ macro defcomp(name, ex)
 
             kws = collectkw(line.args[2].args)
 
-            # Get description and units, if provided
-            description = get(kws, :description, "") # Proposal: change to 'nothing'
-            units = get(kws, :units, "") # Proposal: change to 'nothing'
+            # Get description and unit, if provided
+            description = get(kws, :description, "")
+            unit = get(kws, :unit, "")
 
             if haskey(kws, :index)
                 parameterIndex = kws[:index].args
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by jrising** [comment]
What do you make of the error that this is getting on Linux?
```
$ if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
fatal: git fetch-pack: expected shallow list
```

**Comment by davidanthoff** [comment]
Given that both appveyor and travis have an error in the git clone step, maybe github was down for a short time and they couldn't fetch? Certainly looks like some sort of timeout problem that is not related to your code at all.

I'll look at the PR once I'm done with my lecture!

File: src/Mimi.jl [hunk 1]
```diff
@@ -271,12 +278,25 @@ end
 function connectparameter(m::Model, target_component::Symbol, target_name::Symbol, source_component::Symbol, source_name::Symbol)
     c_target = m.components[target_component]
     c_source = m.components[source_component]
+
+    # Check the units, if provided
+    @assert unitcheck(getmetainfo(m, target_component).parameters[target_name].unit,
+                      getmetainfo(m, source_component).variables[source_name].unit) "Units of $source_component.$source_name do not match $target_component.$target_name."
```
davidanthoff: I think we should not use ``@assert`` here, there is some discussion that those statements might turn into no-ops for higher optimization levels in julia (JuliaLang/julia#10614). Probably safest to just use an ``if`` and then ``throw`` construction here. I would just throw an ``ErrorException`` if things don't match.

src/Mimi.jl
```diff
@@ -271,12 +278,25 @@ end
...
+
     setfield!(c_target.Parameters, target_name, getfield(c_source.Variables, source_name))
     push!(m.parameters_that_are_set, string(target_component) * string(target_name))
     nothing
 end
 
 """
+Default string, string unit check function
+"""
+function unitcheck(one::Union{UTF8String, ASCIIString}, two::Union{UTF8String, ASCIIString})
```
davidanthoff: You can just declare the types of the parameters as ``AbstractString``, no need for a ``Union`` here, I believe

src/Mimi.jl
```diff
@@ -271,12 +278,25 @@ end
...
+    # True if both "", either are "any", or they match
+    return one == "any" || two == "any" || one == two
```
davidanthoff: Could an empty string just mean "any"? I understand why we need something like "any", but I think I would prefer less of a special string...

src/Mimi.jl
```diff
@@ -379,6 +399,20 @@ function getvpd(s)
...
+function collectkw(args::Vector{Any})
+    kws = Dict{Symbol, Any}()
+    for arg in args
+        if isa(arg, Expr) && arg.head == :kw
+            kws[arg.args[1]] = arg.args[2]
+        end
+    end
+
+    kws
+end
```
davidanthoff: Nice, should have done this long ago!

src/Mimi.jl
```diff
@@ -404,17 +438,23 @@ macro defcomp(name, ex)
...
             end
 
-            if any(l->isa(l,Expr) && l.head==:kw && l.args[1]==:index,line.args[2].args)
-                parameterIndex = first(filter(l->isa(l,Expr) && l.head==:kw && l.args[1]==:index,line.args[2].args)).args[2].args
+            kws = collectkw(line.args[2].args)
+
+            # Get description and units, if provided
+            description = get(kws, :description, "") # Proposal: change to 'nothing'
+            units = get(kws, :units, "") # Proposal: change to 'nothing'
+
```
davidanthoff: I actually like empty strings as defaults, so I would leave it like that.

src/metainfo.jl
```diff
@@ -5,15 +5,15 @@ type MetaVariable
     datatype::DataType
     dimensions::Array{Any}
     description::UTF8String
-    unit::UTF8String
+    unit::Any
 end
```
davidanthoff: Given that this only works with strings right now, maybe we can leave this as type string for now? We could make it an ``AbstractString``, or better even just leave it for now?

src/metainfo.jl
```diff
@@ -5,15 +5,15 @@ type MetaVariable
...
 
 type MetaParameter
     name::Symbol
     datatype::DataType
     dimensions::Array{Any}
     description::UTF8String
-    unit::UTF8String
+    unit::Any
```
davidanthoff: Same here.

test/test_units.jl
```diff
@@ -0,0 +1,29 @@
...
+end
+
+m = Model()
+setindex(m, :time, 1)
+foo = addcomponent(m, FooUnits)
+bar = addcomponent(m, BarUnits)
+baz = addcomponent(m, BazUnits)
+
+@test_throws AssertionError bar[:input] = foo[:output]
+baz[:input] = foo[:output]
```
davidanthoff: Maybe also add tests for the "any" cases? Might change of course a bit, if we replace "any" with "" or something else.

davidanthoff:
Alright, I've added some smallish comments. Very nice, thanks a lot!

jrising:
Like the commit message says, I tried to address all of these pieces in light of our conversation.  In particular:
* connectparameter can take an ignoreunits=true optional argument to ignore the units.
* Otherwise, any unit (including "") that does not match will be an error.
* The term 'units' has been changed to 'unit'.

Plus the other things you asked for:
* direct throwing rather than asserting.
* Using AbstractString rather than a Union (I forgot that existed!)
* Switching the type for unit back to UTF8String... for now.

davidanthoff:
Perfect, thanks a lot!

PR #30, opened 2016-04-27
Add clock type and use in run method

davidanthoff:
This is in preparation for a more general API overhaul. For now this should all be considered internal.

Eventually I want to pass a ``Clock`` object to the ``timestep`` function, and then components should use ``gettimestep`` to extract the current t at the top of their ``timestep`` function. Major benefit of such a design would be that we could later add bells and whistles to the ``Clock`` object without breaking the call signature for ``timestep``, i.e. we could add features without breaking old components.

@jrising could you briefly make sure this doesn't break any of your use cases? Right now this should be entirely non-breaking.

### DIFF
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -1,6 +1,7 @@
 module Mimi
 
 include("metainfo.jl")
+include("clock.jl")
 using DataStructures
 using DataFrames
 using Distributions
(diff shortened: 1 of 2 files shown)


jrising:
This is fine with my code.  Thanks!

PR #31, opened 2016-05-04
Rename timestep to run and pass Clock instead of timestep

davidanthoff:
This is breaking change. There are two specific changes: 1) the ``timestep`` function is renamed to ``run`` and 2) instead of passing the time index, I now pass a ``Clock`` instance from which one can extract the current timestep via the ``gettimestep`` function.

Motivation for 1) is that it always bugged me that the function name wasn't a verb. Plus, if we ever change it then now, while we still roughly know who is using this. If you don't like the name, speak now or never ;)

Motivation for 2) is that we can later on pass more information into components without a need to update the call signature again by adding more fields to the ``Clock`` structure.

@jrising, @frankerrickson, any review would be appreciated! I plan to merge this maybe end of the week if I don't hear back. Right now I think there might be one more breaking change in how I handle external parameters, and then I hope to release a v1.0, after which backwards compatibility will be high on the list when we improve Mimi.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # Mimi.jl v.next Release Notes
 * Move type generation from macro to runtime phase (low level dev feature)
+* Rename the timestep function to run_timestep
 
 # Mimi.jl v0.0.3 Release Notes
 * Drop julia 0.3 support
--- a/doc/tutorial.md
+++ b/doc/tutorial.md
@@ -6,7 +6,7 @@ In this example, we will construct a stylized model of the global economy and it
 There are two main steps to creating a component:
 
 * Define the component using ``@defcomp`` where the parameters and variables are listed.
-* Use the timestep function ``timestep(state::component_name, t:Int)`` to set the equations of that component.
+* Use the run_timestep function ``run_timestep(state::component_name, t::Int)`` to set the equations of that component.
 
 Starting with the economy component, each variable and parameter is listed. If either varialbes or parameters have a time-dimension, that must be set with ``(index=[time])``.
 
@@ -25,10 +25,10 @@ using Mimi
 end
 ```
 
-Next, the timestep function must be defined along with the various equations of the ``grosseconomy`` component. In this step, the variables and parameters are linked to this component by ``state`` and must be identified as either a variable or a parameter in each equation. For this example, ``v`` will refer to variables while ``p`` refers to paremeters.
+Next, the run_timestep function must be defined along with the various equations of the ``grosseconomy`` component. In this step, the variables and parameters are linked to this component by ``state`` and must be identified as either a variable or a parameter in each equation. For this example, ``v`` will refer to variables while ``p`` refers to paremeters.
 
 ```julia
-function timestep(state::grosseconomy, t::Int)
+function run_timestep(state::grosseconomy, t::Int)
 	v = state.Variables
 	p = state.Parameters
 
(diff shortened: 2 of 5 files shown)


jrising:
Haven't looked through the code yet, but I have two thoughts.

1. I don't think this needs to be a breaking change.  We can define:
```
run(s::AbstractComponent, clock::Clock) = timestep(s, gettimestep(clock))
```

2. I don't like `run()`, because it does something qualitatively different from the model run method.  In particular, I think components and models could become the same thing in the future, so you could have components that contain other components, and `run` would always mean a whole timeseries.  I would prefer `tick` or `step`.

jrising:
Also, why just pass a Clock object?  Whenever I build ABM systems anyway, I invariably want to pass in the whole model, so that the clock tick function can make changes to it (like adding new agents).  You could pass in a GlobalState object, which would also have the `gettimestep` function.

Aside from those comments, this seems fine.

davidanthoff:
The potential merging of a component and model is a very good point. I guess the only argument in favor of ``run`` is that a lot of similar frameworks seem to use that (HECTOR, ESMF, and I think the whole CESM stuff). But I think your argument is actually stronger, so we should not use ``run``.

Now, ``tick`` and ``step`` are very time-centric, and what I liked about `` run`` was that it would maybe fit nicer with the kind of static model you are building... Maybe ``compute``, or ``execute`` or something? None of these seem great...

Yes, your suggestion would keep backward compatibility... I'm almost tempted though to just break things at this point and then move to a mode where we preserve backwards compat once we hit 1.0. Otherwise we will carry these hacks around forever... People can always decide to stay a bit longer on an old Mimi version if they don't want to update now.

I don't think components should be able to access the whole model object. One of the design principles I have in mind is that components should never take on dependencies on other components, but once a component can access a model that would probably happen.

But, I still like the idea of not passing in a clock as an extra parameter, we could just pass it in as part of the ``state``, and then extract it from that.

jrising:
Quick thought: We could keep `run` and make a more conceptual change to
justify breaking code if we say that that the purpose of `run` is always to
perform computations from t0 to t1.  Then `run` applied to model would do
what it currently does, except that when it calls `run` on the individual
components, it says `run(component, t0::Clock0, t1::Clock1)` or better
yet `run(component, t01::Timespan)`.

 - James

jrising:
How about `update` for the component call?  Then the component state could
include the new time, and the implication would be that the update function
is supposed to finish updating the component to its new state.

- James

davidanthoff:
How about this: for now all I'll do here is rename ``timestep`` to ``run_timestep``, no other changes. I think the whole idea about clocks etc. needs more thought before we can come up with a good design. But I *do* want to get rid of the ``timestep`` name, simply because I hated it from day 1, and having the term "run" appear is good because *so* many models in this space that I've looked at have that as the name of the major routine. Depending models will have to be updated, but that should be one of the last breaking changes before 1.0, and we'll promise backwards compatibility once we have a 1.0 (soon, hopefully).

jrising:
Okay.  I don't like the underscore, given Julia's conventions, but I
understand the reason, and it's just a name anyway.

- James

davidanthoff:
Cool! Technically underscores are ok according to the [julia style guide](http://docs.julialang.org/en/release-0.4/manual/style-guide/#use-naming-conventions-consistent-with-julia-s-base), but in this case there is clearly a better design out there that would use shorter terms, but I think we just haven't found it yet. My main reason for ``run_timestep`` right now is that I don't think we would have a hard time introducing another, more elegant thing later one once we have found one :)

I'll update the PR in a bit.

PR #32, opened 2016-05-26
WIP dependency testing

ckingdon95:


davidanthoff:
Here is another idea: instead of cloning the git repos that have the model code (i.e. fund.jl etc.) you could just download them as a zip file and extract them. That would have two advantages: 1) slightly simpler to specify multiple versions/branches etc. that we want to test, essentially one could just provide a different URL for each of those, 2) we wouldn't require git to be present for this to work, which might be useful given that git won't be bundled with julia 0.5 anymore.

The URLs for downloading a specific branch/tag from github look like this: https://github.com/davidanthoff/ExcelReaders.jl/archive/v0.3.1.zip for example, this would download the version tagged as "v0.3.1". You can also specify a branch name instead. So that is super easy. Unzipping the file should also work, by shelling out to 7z on Windows (bundled with julia) or unzip on any unix based system. Here are two lines that show how that could work: https://github.com/JuliaLang/BinDeps.jl/blob/5163d34ea55e8fa53f4c224f441a1c1917798e55/src/BinDeps.jl#L90, https://github.com/JuliaLang/BinDeps.jl/blob/5163d34ea55e8fa53f4c224f441a1c1917798e55/src/BinDeps.jl#L104.

davidanthoff:
And [davidanthoff/fund.jl](https://github.com/davidanthoff/fund.jl) now has a runtests.jl in the test folder, so that is ready to be tested.

ckingdon95:
Just want to make sure I understand what this would mean: so the test_dependencies.jl would have a list of URLs of branches to loop through, downloading and testing each one. Basically the same as we discussed, but download the zip file of the branch to be tested each time instead of cloning each repository?

davidanthoff:
Yes, exactly.

It would just be a little simpler for starters, because you would just have a list of URLs, instead of a list of tuples with git url + branch name.

test/test_dependencies.jl
```diff
@@ -0,0 +1,47 @@
+using Mimi
+
+#list of URLs of branches of packages to test
+dependencies = ["https://github.com/davidanthoff/fund.jl/archive/master.zip"]
+
+#list of failed tests to build as you go
+errors = []
+
+#make a temporary directory to run the tests in
+run(`mkdir tmp_testing`)
```
davidanthoff: We should probably not shell out for file system operations because this will probably not work cross-platform. Better to use the julia filesystem functions http://docs.julialang.org/en/release-0.4/stdlib/file/.

test/test_dependencies.jl
```diff
@@ -0,0 +1,47 @@
...
+cd("tmp_testing")
```
davidanthoff: Ideally we would not rely on the current path anywhere, but instead use absolute paths everywhere that are inferred from the location of this file. Essentially the idea would be to use something like this ``tmp_path = joinpath(dirname(@__FILE__),"tmp_testing")``. Not urgent, we can discuss tomorrow.

test/test_dependencies.jl
```diff
@@ -0,0 +1,47 @@
...
+
+#loop through each dependent package
+for url in dependencies
+  #download the package
+  run(`wget $url`)
```
davidanthoff: This won't work on Windows. You can try to use [download](http://docs.julialang.org/en/release-0.4/stdlib/file/?highlight=download#Base.download) instead.

test/test_dependencies.jl
```diff
@@ -0,0 +1,47 @@
...
+  zip_name = chomp(readall(`basename $url`))
```
davidanthoff: Maybe https://github.com/JuliaWeb/URIParser.jl could be used here so that this works cross-platform?

test/test_dependencies.jl
```diff
@@ -0,0 +1,47 @@
...
+  run(`unzip -x $zip_name`)
```
davidanthoff: Two options here: there might be a package that does this in a cross-platform way. Otherwise you have to use ``7z`` on windows and ``unzip`` on unix/mac.

test/test_dependencies.jl
```diff
@@ -0,0 +1,47 @@
...
+  run(`rm $zip_name`)
```
davidanthoff: For all of these file system ops, just use the julia functions I linked to above.

test/test_dependencies.jl
```diff
@@ -0,0 +1,47 @@
...
+  package_name = chomp(readall(`ls`))
+  #test the package
+  try
+    Pkg.test(package_name)
```
davidanthoff: This won't work because the git repos like fund.jl are not actually packages. Here you'll actually have to shell out to julia itself and pass the name of the script that should be run to julia.

test/test_dependencies.jl
```diff
@@ -0,0 +1,47 @@
...
+  println("1 test failed:")
+else
+  println("$num_errors tests failed:")
+end
+
+for item in errors
+  package_name, error = item
+  println("error in $package_name:")
+  println(error)
+end
```
davidanthoff: We somehow have to fail properly here if there are errors. Not sure whether we just throw an error, or whether there is some better way to hook into julia's unit testing framework here. The important thing is that we somehow need to signal to travis that there was an error and that travis should show the build as failed.

davidanthoff:
This all looks good, the structure is exactly right! I've added a lot of small comments that should be relatively easy to fix, most of them are about cross-platform issues.

davidanthoff:
Ah, and you should add this new test to https://github.com/davidanthoff/Mimi.jl/blob/master/test/runtests.jl#L4, so that travis and appveyor will actually run this. That would also show you cross-platform problems.

test/test_dependencies.jl
davidanthoff: You can try this https://github.com/samoconnor/InfoZIP.jl#unzip.

ckingdon95:
Okay great, thanks for the helpful links! I will try to make the changes tonight.

test/test_dependencies.jl
```diff
@@ -0,0 +1,54 @@
+using Mimi
+
+Pkg.add("InfoZIP")
```
davidanthoff: Package dependencies are specified in a REQUIRE file in the test folder. Just add a text file named REQUIRE to the test folder that has one line with the content ``InfoZIP`` and the package manager will install the InfoZIP package automatically before it tries to run the tests. Then add a ``using InfoZIP`` line here.

test/test_dependencies.jl
```diff
@@ -0,0 +1,54 @@
...
+
+#list of URLs of branches of packages to test
+dependencies = [
+"https://github.com/davidanthoff/fund.jl/archive/master.zip"
+]
+
+#list of failed tests to build as you go
+errors = []
+#make a temporary directory to run the tests in
+tmp_path = joinpath(dirname(@__FILE__),"tmp_testing/")
```
davidanthoff: Minor nitpick, I think you can leave out the ``/`` at the end, that is platform specific and ``joinpath`` should take care of that in any case.

test/test_dependencies.jl
```diff
@@ -0,0 +1,54 @@
...
+mkdir(tmp_path)
+cd(tmp_path)
+
+#loop through each dependent package
+for url in dependencies
+  #download the package
+  download(url)
+  zip_name = chomp(basename(url))
```
davidanthoff: Does ``basename`` work here? I thought it it meant for filenames, not URLs? I might be wrong.

test/test_dependencies.jl
```diff
@@ -0,0 +1,54 @@
...
+  unzip(zip_name, tmp_path)
+  rm(string(tmp_path, zip_name))
+  package_name = chomp(readall(`ls`))
```
davidanthoff: As we just discussed, ``readdir`` should do the trick.

test/test_dependencies.jl
```diff
@@ -0,0 +1,54 @@
...
+  #test the package
+  try
+    include(string(tmp_path, package_name, "/test/runtests.jl"))
```
davidanthoff: Use ``run(`julia filename.jl`)`` etc.

test/test_dependencies.jl
```diff
@@ -0,0 +1,54 @@
...
+rm(tmp_path, recursive=true)
+
+#report the errors that occurred
+num_errors = length(errors)
+if num_errors == 1
+  error_message = "1 test failed \n:
+  "
+else
+  error_message = "$num_errors tests failed: \n"
+end
```
davidanthoff: I would just do ``error_message = "Failed tests: $num_errors"``.

test/test_dependencies.jl
```diff
@@ -0,0 +1,54 @@
...
+
+for item in errors
+  package_name, error = item
```
davidanthoff: I think you can unpack the tuple in the for loop already:
````julia
for (package_name, error) in errors
````

File: test/test_dependencies.jl [hunk 16]
```diff
@@ -0,0 +1,54 @@
...
+  error_message = string(error_message, "error in $package_name:", error, "\n")
+end
+
+if num_errors > 0
+  throw(Exception: error_message)
```
davidanthoff: As discussed, use ``error(error_message)`` here.

REQUIRE
```diff
@@ -3,3 +3,4 @@ DataStructures
 Distributions
 DataFrames
 Docile
+InfoZIP
```
ckingdon95: sorry I think is wrong, ignore this commit-- the next commit has the set up I believe you wanted

test/test_dependencies.jl
ckingdon95: Yes it works for URL paths as well!

ckingdon95:
I'm not sure why it can't find the "unzip" function anymore; I've included "using InfoZIP" at the top of the file and in the REQUIRE file. Do you know what the problem is?

davidanthoff:
Try to add ``ZipFile`` to the test/REQUIRE in the first line. Plus, ``unzip`` is not exported (http://docs.julialang.org/en/release-0.4/manual/modules/) so you need to call it with ``InfoZIP.unzip(blabla)``.

See also samoconnor/InfoZIP.jl#2, at the end of the day I think the InfoZIP package should either just properly declare a dependency on ``ZipFile``, or the README in InfoZIP should explain more clearly how one can use the ``unzip`` function. Let's see whether my PR gets merged, if not, we can try to improve the README there.

davidanthoff:
And see also samoconnor/InfoZIP.jl#3.

tkelman:
I warned you about this somewhere, but testing against a moving target master of a different resource has made your package start failing its tests. http://pkg.julialang.org/logs/Mimi_0.5.log

You need to be using a specific sha or tag so you don't cause tests to fail on past versions of the julia package.

davidanthoff:
Yes, good point. I'm tagging a bug fix release right now.

samoconnor:
Hi @davidanthoff, thx for suggesting InfoZIP.jl README update: https://github.com/samoconnor/InfoZIP.jl/commit/17f3ba3faa8836caf30a1ac41f3e611736deef15

PR #37, opened 2016-07-18
WIP Model graph

ckingdon95:


src/Mimi.jl
```diff
@@ -225,6 +228,7 @@ function addcomponent(m::Model, t, name::Symbol;before=nothing,after=nothing)
     end
 
     ComponentReference(m, name)
+    add_node(m.model_graph, name)
 end
```
davidanthoff: This needs to be above the previous line. The last line of a function is its return value (unless you specify a return value explicitly with ``return x``), so this line changes the return value of the function (and actually that breaks the units tests.

src/Mimi.jl
```diff
@@ -301,6 +305,8 @@ function connectparameter(m::Model, target_component::Symbol, target_name::Symbo
 
     setfield!(c_target.Parameters, target_name, getfield(c_source.Variables, source_name))
     push!(m.parameters_that_are_set, string(target_component) * string(target_name))
+    e=edge(target_name, target_component, source_component)
```
davidanthoff: ``source-Name`` should also be part of the edge.

src/graph_API.jl
```diff
@@ -0,0 +1,90 @@
+type edge
```
davidanthoff: Typenames typically start with capital letters in julia.

src/graph_API.jl
```diff
@@ -0,0 +1,90 @@
...
+  # param_node: the "in" node (listed first in connectparameter)
+  # var_node: the "out" node (listed second in connectparameter)
+  name::Symbol
```
davidanthoff: Connections can go from a var to a par with different names, so you'll need two fields here for the source and target name.

src/graph_API.jl
```diff
@@ -0,0 +1,90 @@
...
+  param_node::Symbol #name of component
+  var_node::Symbol #name of component
+end
+
+type ModelGraph
+  edges
+  nodes
+  size
+end
```
davidanthoff: All three fields here should be typed.

src/graph_API.jl
```diff
@@ -0,0 +1,90 @@
...
+ModelGraph(e,n) = ModelGraph(e, n, length(n))
+ModelGraph() = ModelGraph([],[],0)
+
+function add_edge(g::ModelGraph, e::edge)
+  @assert e.param_node in g.nodes && e.var_node in g.nodes ["Tried to add an edge between vertices not in the graph"]
+  append!(g.edges, [e])
```
davidanthoff: Why the ``[]`` around the ``e``? You want to add one element, not a vector of one element, right?

src/graph_API.jl
```diff
@@ -0,0 +1,90 @@
...
+    o_edges = get_edges(g,c,"outgoing")
+    println(c)
+    println("  incoming parameters:")
+    [println("    - ",e.name," from ",e.var_node) for e in i_edges]
+    println("  outgoing variables:")
+    [println("    - ",e.name," from ",e.param_node) for e in o_edges]
+  end
+end
+
+function string_representation(g::ModelGraph)
```
davidanthoff: What is the purpose of this function?

src/graph_API.jl
```diff
@@ -0,0 +1,90 @@
...
+    end
+    rep = string(rep, "  outgoing variables:", "\n")
+    for e in o_edges
+      rep = string(rep, "    - ",e.name," from ",e.var_node,"\n")
+    end
+  end
+  return rep
+end
+
+function get_edges(g::ModelGraph, component::Symbol, which::String)
```
davidanthoff: Make the ``which`` parameter of type ``Symbol``. And maybe add a default value?

davidanthoff:
This looks good! I have lots of small inline nitpicks, but the general structure seems all great.

Apart from the small inline comments, two other things:
- Try to run the existing unit tests of the package before you commit. Just execute ``Pkg.test("Mimi")`` on the julia REPL, that should do the trick. If you forget, you should still see those errors here in the CI test results (i.e. both the travis and appveyor build failed right now for this PR).
- We should get into a habit of adding units tests for new functionality right away. I'm more than guilty of not doing this in the past, but lets try to do so in the future. Do you know the general idea of how that works? If not, we can discuss on Wed.

src/Mimi.jl
ckingdon95: Yes, I am going to make the ComponentInstanceInfo type that we discussed, so that nodes will end up holding all of the component's information (right now they are just symbols)

src/graph_API.jl
ckingdon95: Yes, just the length of nodes. Wasn't sure if it would be faster to keep it stored as a variable, but I imagine that asking for the length of an array is not a very costly operation, so I can delete this field if you want

src/graph_API.jl
ckingdon95: From messing around with commands in a julia console I thought this was how `append!` works, not sure though. If I do it without the brackets I get a method error, saying that both arguments must be arrays.

src/graph_API.jl
ckingdon95: The sole purpose of this function was just for me to be able to see if I was constructing the graph correctly; I just wrote it as an ad hoc way to loosely test functionality and (sort of) visualize connections between components. (this function will get deleted eventually)

ckingdon95:
There are still several things I plan on adjusting by Wednesday, including writing some unit tests. Also, would you prefer if I just only commit my changes locally (and hold off on pushing to github) until whatever I'm working on is more or less finalized? Or would you prefer to see code (like this commit) as I'm working on it, even if there are pieces that aren't completed and/or do not pass the package tests?

src/graph_API.jl
davidanthoff: Yes, there should be no perf implication of using ``length`` here.

src/graph_API.jl
davidanthoff: Ah, you are right! But, you can use ``push!`` instead to only add one element. That should be better because it will save on array allocation.

davidanthoff:
I think it is good to have a PR open from the start so that we can have some back-and-forth even while you work on the stuff, certainly no harm in that. Maybe start the title with "WIP: " (work in progress) to indicate that no-one should merge at this point.

src/graph_API.jl
ckingdon95: Ah got it, thanks!

ckingdon95:
That sounds great, that structure makes a lot of sense!

src/Mimi.jl
```diff
@@ -2,14 +2,15 @@ module Mimi
 
 include("metainfo.jl")
 include("clock.jl")
+include("graph_API.jl")
```
davidanthoff: Delete

src/Mimi.jl
```diff
@@ -2,14 +2,15 @@ module Mimi
...
 using DataStructures
 using DataFrames
 using Distributions
 
 export
     ComponentState, run_timestep, run, @defcomp, Model, setindex, addcomponent, setparameter,
     connectparameter, setleftoverparameters, getvariable, adder, MarginalModel, getindex,
-    getdataframe, components, variables, setbestguess, setrandom, getvpd, unitcheck
+    getdataframe, components, variables, setbestguess, setrandom, getvpd, unitcheck, print_graph
```
davidanthoff: Don't export anything, instead we overwrite the ``show`` method as discussed yesterday.

src/Mimi.jl
```diff
@@ -121,13 +122,28 @@ end
 function setrandom(p::CertainArrayParameter)
 end
 
+type ComponentInstanceInfo
+  name::Symbol
+  component_type #the type here is a Type? or ComponentState or something else?
```
davidanthoff: I think ``DataType``...

src/Mimi.jl
```diff
@@ -121,13 +122,28 @@ end
...
+
 type Model
     indices_counts::Dict{Symbol,Int}
     indices_values::Dict{Symbol,Vector{Any}}
     components::OrderedDict{Symbol,ComponentState}
     parameters_that_are_set::Set{UTF8String}
     parameters::Dict{Symbol,Parameter}
     numberType::DataType
+    #model_graph::ModelGraph
+    edges::Array{ParameterVariableConnection, 1}
```
davidanthoff: Can we use a more specific name here, e.g. ``connections`` or something like that?

src/Mimi.jl
```diff
@@ -121,13 +122,28 @@ end
...
+    nodes::Dict{Symbol, ComponentInstanceInfo}
```
davidanthoff: Same here. I'd actually like to call it ``components``, but that is taken right now. Maybe call it ``components2`` for now, and once we have merged this PR we'll rename the original ``components`` to something else and then you can use ``components`` here. But lets do that later.

src/Mimi.jl
```diff
@@ -137,6 +153,8 @@ type Model
         m.parameters_that_are_set = Set{UTF8String}()
         m.parameters = Dict{Symbol, Parameter}()
         m.numberType = numberType
+        m.edges = []
```
davidanthoff: Does this work? I would think ``[]`` creates an array of type ``Array{Any,1}``. ``Array(ParameterVariableConnection, 0)`` Should create a vector with the right types and 0 elements.

src/Mimi.jl
```diff
@@ -224,6 +242,11 @@ function addcomponent(m::Model, t, name::Symbol;before=nothing,after=nothing)
         m.components[name] = comp
     end
 
+    #add_node(m.model_graph, name)
+    this = ComponentInstanceInfo(name, t) #build node instance
+    #push!(m.nodes, this) #add to model's list of nodes
+    m.nodes[name] = this
```
davidanthoff: Can you delete the lines with comment signs?

src/Mimi.jl
```diff
@@ -571,6 +601,62 @@ function run_timestep(s::adder, t::Int)
...
+  end
+end
+
+function get_connections(m::Model, c::ComponentInstanceInfo, which::Symbol)
+  return get_connections(m, c.name, which)
+end
+
+function get_connections(m::Model, component_name::Symbol, which::Symbol)
+  #which = uppercase(which)
+  if which==:ALL
```
davidanthoff: Can you make these values lowercase, i.e. ``:all`` etc. For some reason that seems the julia convention. Btw., I think there might also be an enum type in julia that could be used here. I haven't really looked into that...

src/Mimi.jl
```diff
@@ -571,6 +601,62 @@ function run_timestep(s::adder, t::Int)
...
+    function bool(e::ParameterVariableConnection)
```
davidanthoff: This is where you want to use an anonymous function, right? That would be better here.

src/Mimi.jl
```diff
@@ -571,6 +601,62 @@ function run_timestep(s::adder, t::Int)
...
+  elseif which==:INCOMING
+    function bool(e::ParameterVariableConnection)
+      return e.target_component_name==component_name
+    end
+  elseif which==:OUTGOING
+    function bool(e::ParameterVariableConnection)
+      return e.source_component_name==component_name
+    end
+  else
+    @assert false ["Invalid parameter to the 'which' argument; must be 'all' or 'incoming' or 'outgoing'."]
```
davidanthoff: You can just write ``error("Whatever you want your error message to be")`` here.

src/Mimi.jl
```diff
@@ -571,6 +601,62 @@ function run_timestep(s::adder, t::Int)
...
+  end
+
+  lst=[]
```
davidanthoff: Create a typed array here (see above).

src/Mimi.jl
```diff
@@ -571,6 +601,62 @@ function run_timestep(s::adder, t::Int)
...
+  for e in m.edges
+    if bool(e)
+      push!(lst, e)
+    end
+  end
```
davidanthoff: You can use the ``filter`` [function](http://docs.julialang.org/en/release-0.4/stdlib/collections/?highlight=filter#Base.filter) here, that can replace the whole for loop.

ckingdon95:
Ready for you to review the code, then I will write a few more tests (including a test for the show function). Then I will start working on the new build function.

src/Mimi.jl
```diff
@@ -215,17 +232,24 @@ function addcomponent(m::Model, t, name::Symbol=Symbol(string(t)); before=nothin
 
     if before!=nothing
         newcomponents = OrderedDict{Symbol,ComponentState}()
+        newcomponents2 = OrderedDict{Symbol, ComponentInstanceInfo}()
         for i in keys(m.components)
             if i==before
                 newcomponents[name] = comp
+                this = ComponentInstanceInfo(name, t)
```
davidanthoff: Maybe use a different name here than ``this``. That term is reserved in lots of programming languages, and who knows, maybe the julia people want to use it at some point as well.

src/Mimi.jl
```diff
@@ -571,6 +599,50 @@ function run_timestep(s::adder, t::Int)
     v.output[t] = p.input[t] + p.add[t]
 end
 
+#Begin Graph Functionality section
+
+function show(io::IO, m::Model)
+    println(io, "showing model component connections:")
+    i=1
+    for c in keys(m.components2)
```
davidanthoff: You can use the [``enumerate``](http://docs.julialang.org/en/release-0.4/stdlib/collections/?highlight=enumerate#Base.enumerate)  function to give you the ``c`` and the ``i`` here.

davidanthoff:
This looks great! The appveyor failures are caused by upstream problems so we can ignore those. The failed travis julia 0.4 build timed out, so we can ignore that as well, but there seems to be some problem with the julia 0.5 builds on travis that I actually don't understand.

I would suggest you start to write the remaining tests, the overall design here is definitely correct, and then we'll try to figure out what this error on julia 0.5 means.

src/Mimi.jl
```diff
@@ -571,6 +599,50 @@ function run_timestep(s::adder, t::Int)
...
+    end
+end
+
+function get_connections(m::Model, c::ComponentInstanceInfo, which::Symbol)
+    return get_connections(m, c.name, which)
+end
+
+function get_connections(m::Model, component_name::Symbol, which::Symbol)
+    if which==:all
+        bool(e) = e.source_component_name==component_name || e.target_component_name==component_name
```
davidanthoff: For some reason this causes troubles on julia 0.5. I don't understand why, but the following should solve the problem: instead of defining a regular function here, create an anonymous function and assign it to a variable. Maybe, just to be save, also don't use ``bool``, that sounds very much like something that julia might use at some point. This line would look like this:
````julia
f = e-> e.source_component_name==component_name || e.target_component_name==component_name
````
And then you pass ``f`` to ``filter``.

PR #38, opened 2016-07-19
Add conveniences for OptiMimi

jrising:
Both of these should make no difference for normal usage, but make things more convenient for a couple of use cases.  There are two changes:

* Combined the two `addcomponent` calls.  The only advantage for this is if you have another function which is calling `addcomponent` which could take a `name` parameter, and then can just pass it in.
* If the model uses a type other than `Float64`, it automatically does conversions.  So you can always call `setparameter` the same way, no matter what the number type of the model is.

### DIFF
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -195,12 +195,12 @@ end
 """
 Add a component to a model.
 """
-function addcomponent(m::Model, t, name::Symbol=:nothing; before=nothing,after=nothing)
+function addcomponent(m::Model, t, name::Union{Symbol,Void}=nothing; before=nothing,after=nothing)
     if before!=nothing && after!=nothing
         error("Can only specify before or after parameter")
     end
 
-    if name == :nothing
+    if name == nothing
         name = symbol(string(t))
     end
 
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -192,9 +192,16 @@ function setindex{T}(m::Model, name::Symbol, values::Vector{T})
     nothing
 end
 
-function addcomponent(m::Model, t, name::Symbol;before=nothing,after=nothing)
+"""
+Add a component to a model.
+"""
+function addcomponent(m::Model, t, name::Symbol=:nothing; before=nothing,after=nothing)
     if before!=nothing && after!=nothing
         error("Can only specify before or after parameter")
+    end
+
+    if name == :nothing
+        name = symbol(string(t))
     end
 
     comp = t(m.numberType, m.indices_counts)
@@ -216,15 +223,6 @@ function addcomponent(m::Model, t, name::Symbol;before=nothing,after=nothing)
 
     ComponentReference(m, name)
 end
-
-function addcomponent(m::Model, t;before=nothing,after=nothing)
-    addcomponent(m,t,symbol(string(t)),before=before,after=after)
-end
-
-"""
-Add a component to a model.
-"""
-addcomponent
 
 """
 Set the parameter of a component in a model to a given value.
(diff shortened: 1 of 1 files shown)


jrising:
The CI failure seems to come from a previous commit:
```
ERROR: LoadError: MethodError: `mode` has no method matching mode(::Distributions.Truncated{Distributions.Gamma{Float64},Distributions.Continuous})
```

src/Mimi.jl
```diff
@@ -192,11 +192,18 @@ function setindex{T}(m::Model, name::Symbol, values::Vector{T})
     nothing
 end
 
-function addcomponent(m::Model, t, name::Symbol;before=nothing,after=nothing)
+"""
+Add a component to a model.
+"""
+function addcomponent(m::Model, t, name::Symbol=:nothing; before=nothing,after=nothing)
```
davidanthoff: Can you make the name argument ``name::Union{Symbol,Void}=nothing``? I'm not a fan of sentinel values. My suggestion will make the function type-instable, but I think we can ignore performance implications here.

src/Mimi.jl
```diff
@@ -269,6 +267,11 @@ function addparameter(m::Model, name::Symbol, value)
             p = UncertainArrayParameter(value)
             m.parameters[symbol(lowercase(string(name)))] = p
         else
+            if !(typeof(value) <: Array{m.numberType})
+                # E.g., if model takes Number and given Float64, convert it
+                value = convert(Array{m.numberType}, value)
+            end
```
davidanthoff: This is good for now, but hopefully we won't need it anymore once we generate the types just before the ``run`` call because then we can just make the type of the parameter match. But given that that is all future talk, this is really super helpful now.

davidanthoff:
Thanks!

Yes, the CI failure is I believe related to some change in an upstream package that breaks fund.jl, and now breaks our tests... I'm trying to fix that right now.

Generally, could you try to split PRs like this into two? The two changes don't really have anything to do with each other, and then we can merge individually.

jrising:
I made the `Union{}` change.  This is definitely a case where we can sacrifice a minuscule amount of performance, I think.

If you want, I can split this into two pull requests.  They're both motivated by the same change on my end: trying to make no differences in how a component is written for a simulated model or an optimization.

src/Mimi.jl
```diff
@@ -192,11 +192,18 @@ function setindex{T}(m::Model, name::Symbol, values::Vector{T})
     nothing
 end
 
-function addcomponent(m::Model, t, name::Symbol;before=nothing,after=nothing)
+"""
+Add a component to a model.
+"""
+function addcomponent(m::Model, t, name::Union{Symbol,Void}=nothing; before=nothing,after=nothing)
```
davidanthoff: I found another way for this that is way more elegant and type-stable (not that that matters here)! Try
````julia
function addcomponent(m::Model, t, name::Symbol=Symbol(string(t)); before=nothing,after=nothing)
````
and then you can delete the below ``if`` clause entirely. Sorry for the back and forth, but I just discovered that this works myself.

Oh, and the ``symbol`` function is deprecated and one should use ``Symbol`` instead starting with julia 0.4.

davidanthoff:
> If you want, I can split this into two pull requests. They're both motivated by the same change on my end: trying to make no differences in how a component is written for a simulated model or an optimization.

No, don't worry, was more meant for future PRs, we'll keep this as is.

jrising:
I wondered if that kind of syntax works.  Cool!  Made the change.

davidanthoff:
Thanks!

PR #42, opened 2016-08-17
WIP Model build

ckingdon95:


ckingdon95:
@davidanthoff if you have time, would you mind looking at the most recent error that's happening in test_references? the error is actually on line 516 of Mimi in the build function; it's complaining that the component doesn't have a field for the field we are trying to set.

ckingdon95:
also, another potentially related question, on line 27 of test_references, which function does that syntax invoke? does that directly translate to setparameter(m, foo, input, 3.14) ?

davidanthoff:
``builtComponents`` holds instances of the type that implements the component, so e.g. ``FooImpl``. That type has only three fields: ``Parameters``, ``Variables`` and ``Dimensions``. The type of the field ``Parameters`` is ``FooParameters``, and that is the type that actually has fields for each of the component parameters. So you need to change line 517 to this:
````julia
setfield!(builtComponents[x.component_name].Parameters, x.param_name, param.value)
````
Same is true for line 519.

davidanthoff:
> also, another potentially related question, on line 27 of test_references, which function does that syntax invoke? does that directly translate to setparameter(m, foo, input, 3.14) ?

Yes. That syntax calls the function ``setindex!`` on line 23 in references.jl and you can see how it calls ``setparameter`` there.

ckingdon95:
I think my re-write is complete. It's still not passing tests though because Fund tries to use the setbestguess function. I still need to go through and clean up the code before merging the PR (mostly just deleting comments, renaming some variables, etc.)

davidanthoff:
I merged it! There are a few more commits on master that I added. Some are minor bug fixes, but the bigger change I had was around how to handle uninitialized model instances. Instead of having a special ``null`` value I changed that to use a ``Nullable`` type instead. Take a look at the julia documentation how that one works!

PR #43, opened 2016-08-24
Skip model dependencies except on Travis CI

jrising:
Removed the test_dependencies.jl test from runtests, but added it to the travis script.

Note that I cleaned out the `if [ -f test/runtests.jl ]; then` part of the travis script.  If runtests.jl isn't there, that's a pretty big problem too.


davidanthoff:
Thanks!

PR #61, opened 2016-10-11
Fix bug that surfaced in setleftoverparameters

davidanthoff:
Instance names of components in a model can never be resolved in the global meta component namespace, this looks things up with the component type name instead, which is the right case.

@ckingdon95 does my PR here look right to you? We should also add a test that catches this error.


ckingdon95:
Yes that makes sense, sorry about that! But why is it component_type.name.name and not just component_type.name ?

davidanthoff:
I believe ``component_type.name`` returns an instance of a type that then has a number of fields, and if we want to get just the symbol for the name we need to use the syntax that I used.

PR #62, opened 2016-10-12
Drop julia 0.4 support

davidanthoff:


REQUIRE
```diff
@@ -1,6 +1,5 @@
-julia 0.4
+julia 0.5
 DataStructures
 Distributions
 DataFrames
 Documenter
-Compat 0.8.6
```
tkelman: you still have several other uses of Compat

davidanthoff: Thanks, fixed in #90.

Issue #53, opened 2016-10-10
Connections between scalar variables and parameters don't work


davidanthoff:

ckingdon95:
I am working on this with pull request #63

ckingdon95:
This issue has been resolved!

PR #63, opened 2016-10-15
WIP scalar parameter/variable connections

ckingdon95:
addresses issue #53 on anthofflab/Mimi.jl

### DIFF
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -32,10 +32,6 @@ type ComponentInstanceInfo
     component_type::DataType
 end
 
-type ModelInstance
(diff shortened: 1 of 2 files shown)


ckingdon95:
@davidanthoff 
I'm currently passing my tests locally but they are failing here. The error is:

Test threw an exception of type BoundsError
Expression: (m[:A,:varA])[t] == 1
BoundsError: attempt to access 1-element Array{Float64,1} at index [2]


but varA should be a full array. I'm not sure how to go about debugging this since on my machine varA is an array with a value for every timestep. Do you know why this could be happening?

Issue #74, opened 2016-11-21
Move marginal model function to new file 


girishbalaji:

davidanthoff:
Move types and functions into file. Add tests.

PR #77, opened 2016-12-07
WIP marginal model file

ckingdon95:


ckingdon95:
addressing issue #74

ckingdon95:
This is basically done, but I couldn't put the `include("marginalmodel.jl")` at the top of Mimi.jl, because the code in the file has to come after Model is defined. (so right now it's at line 105). Is there a work around for this that you know of? @davidanthoff

davidanthoff:
I think most packages get around that by having the main julia file (i.e. ``Mimi.jl``) just be something like
````julia
module Mimi
include("bla")
include("bla")
end
````
and then they move all real code into different files. That way you can order in any way you want. I think we should do that soonish, but as part of another new PR that does just that. @ckingdon95 can you open a new issue for that so that we don't forget? And if someone wants to tackle it, go ahead.

For this PR, just put the ``include`` somewhere in the middle.

Issue #73, opened 2016-11-21
Move defcomp addr code to a new file


girishbalaji:

davidanthoff:
Plus add tests.

PR #78, opened 2016-12-07
WIP defcomp adder file

ckingdon95:
addressing issue #73

### DIFF
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -141,7 +141,7 @@ end
 function getmetainfo(m::Model, componentname::Symbol)
     meta = metainfo.getallcomps()
     meta_module_name = Symbol(m.components2[componentname].component_type.name.module)
-    meta_component_name = Symbol(m.components2[componentname].component_type)
+    meta_component_name = m.components2[componentname].component_type.name.name
     return meta[(meta_module_name, meta_component_name)]
 end
 
@@ -174,7 +174,7 @@ end
 """
 Add a component to a model.
 """
-function addcomponent(m::Model, t, name::Symbol=Symbol(string(t)); before=nothing,after=nothing)
+function addcomponent(m::Model, t, name::Symbol=t.name.name; before=nothing,after=nothing)
     if before!=nothing && after!=nothing
         error("Can only specify before or after parameter")
     end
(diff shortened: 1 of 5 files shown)


src/Mimi.jl
```diff
@@ -216,14 +216,15 @@ function setparameter(m::Model, component::Symbol, name::Symbol, value)
 end
 
 function checklabels(m::Model, component::Symbol, name::Symbol, parametername::Symbol, p::ArrayModelParameter)
-    if !(eltype(p.values) <: getmetainfo(m, component).parameters[parametername].datatype)
+    meta_comp_name = Symbol(string(m.components2[component].component_type))
+    if !(eltype(p.values) <: getmetainfo(m, meta_comp_name).parameters[parametername].datatype)
```
davidanthoff: I think the error is here. I think you need to call ``getmetainfo`` with the name of the component in the model, not the type of the component. But the way you construct ``meta_comp_name``, it will be the name of the type. I think you just need to change this to ``getmetainfo(m, component)``, and it should work?

davidanthoff: Actually, there is also a bug in ``getmetainfo``. It should be ``meta_component_name = Symbol(m.components2[componentname].component_type.name.name)`` there, not what it is right now.

And then you'll also have to fix the other calls to ``getametainfo`` in ``checklabels``.

ckingdon95: So I actually just added that line yesterday when I was trying to figure out the problem. I'll take it out now, but the error still happens when trying to look up the "adder" component in getmetainfo.

ckingdon95: The error is slightly different depending on whether or not I give the adder component a different name.

When I do:
`addcomponent(model, adder)`
`setparameter(model, :adder, :input, x)`

The error is "KeyError: key :adder not found"

Whereas when I do 

`addcomponent(model, adder, :compA)`
`setparameter(model, :compA, :input, x)`

The error is "KeyError: key (:Mimi,Symbol("Mimi.adder")) not found"

ckingdon95: The problem seems to be that the name "adder" evaluates to Mimi.adder. So when we add the component to the model, what it tries to add is Mimi.adder, which isn't quite what we want. But I don't fully understand why the name "adder" isn't associated with the ComponentState that was created by the `defcomp adder`, like any other component normally is.

davidanthoff:
Great, thanks!

Issue #58, opened 2016-10-10
Test arguments


davidanthoff:
Generally we need to make sure that we validate the input arguments for each public API method before we start the actual work.

davidanthoff:
List of functions:
- [ ] ``run``: check that model has >0 components, check that there are no unset parameters, check that time index is set.
- [ ] ``addcomponent``: check ``t``, check whether component with same name exists already, check whether ``before`` and ``after`` components actually exist.
- [ ] ``setparameter``: Lots
- [ ] ``connectparameter``: Lots
- [ ] ``getindex``, check whether component/var combo exists
- [ ] ``getdataframe`` same
- [ ] ``addparameter``, check for duplicates
- [ ] ``plot`` Check for existence of compos/vars, but might be covered by previous checks.

girishbalaji:
I'm slightly confused as to what checks to add for setparameter and the add and connect parameters. I'm not sure what would be different between the three as well

PR #81, opened 2016-12-12
WIP [To Close] added num component checking

girishbalaji:
#58 Adds check for num components. Will complete the rest with future commits

### DIFF
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -477,7 +477,8 @@ end
 
 function run(mi::ModelInstance, ntimesteps, indices_counts)
     if length(mi.components) == 0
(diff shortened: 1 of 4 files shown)


girishbalaji:
I'm not sure if the Warnings that I'm receiving are causing the failing build. I ran julia runtests.jl locally; everything passes (including the test_arguments file I added) but when it reaches test_tools.jl I get several 'WARNING: Method definition ...  overwritten' issues. I believe I had the same issue on test_arguments which I fixed by creating unique variable names.

davidanthoff:
Next step is to resolve merge conflict.

PR #85, opened 2017-02-10
testing for local dependent packages

ckingdon95:


REQUIRE
```diff
@@ -6,4 +6,4 @@ Documenter 0.7.1
 NamedArrays 0.5.3
 Plots 0.9.4
 JSON
-
+InfoZIP
```
davidanthoff: I'd prefer if this dependency was not listed here. For normal use of Mimi one doesn't need this package, really just for testing, which is probably a very small subset of people that will install Mimi. The previous version had this ``Pkg.installed("ZipFile")`` code at the top of ``test_dependencies.jl``, could you just add that back in but modify it to install InfoZIP if needed?

test/test_dependencies.jl
```diff
@@ -1,76 +1,64 @@
-if Pkg.installed("ZipFile") == nothing
-    Pkg.add("ZipFile")
-end
```
davidanthoff: Reenable that, but for InfoZIP.

test/test_dependencies.jl
```diff
@@ -1,76 +1,64 @@
...
-    run(`julia $process`)
-  catch e
-    append!(errors, [(package_name, e)])
-  end
-  #delete current package before testing next one
-  rm(joinpath(tmp_path, package_name), recursive=true)
-end
+    #loop through each dependent package
+    for d in dependencies
+        if  isa_url(d) #what's the best way to know if it's a url?
```
davidanthoff: Can you remove the comment here? That is sorted out now.

test/test_dependencies.jl
```diff
@@ -1,76 +1,64 @@
...
+            #find the name of the unzipped package (this only works if the zip archive only has one directory, the package)
+            package_name = readdir(tmp_path)[1]
+            process = string(tmp_path, package_name, "/test/runtests.jl")
+        else
+            package_name = basename(d)
+            process = string(d, "/test/runtests.jl")
+        end
+        #test the package
+        try
+            run(`julia $process`)
```
tkelman: `$JULIA_HOME/julia $process`

davidanthoff: Actually surprising that this worked so far...

PR #86, opened 2017-02-10
new indexing and dimensions functions

ckingdon95:


jrising:
This works fine for me!  Thanks.

jrising:
Actually, I didn't see that you rolled in the other issue about parameter and variable dimensions too.  A couple of thoughts:
 - Do we need to have a separate function for parameters and variables?  The parameter-variable distinction is useful elsewhere, but maybe we should just have the function be `getdimensions(model, component, name)`, able to take either.
 - Are we calling these indexes or dimensions?  It's currently inconsistent within this one PR.
 - Julia convention is to avoid underscores unless essential.

davidanthoff:
@jrising We actually were just working on combining the variable and parameter case while you were submitting the issue ;)

@ckingdon95 I had another look at naming, and I think @jrising is right, we should not use underscores, mainly because we also don't use them for e.g. ``getdataframe`` etc. And on "index" vs "dimension" I also agree, we should use only one terminology, and it should be "index" because that is also what we use in the ``@defcomp`` macro. Can you make those two renames?

src/Mimi.jl
```diff
@@ -379,6 +380,28 @@ function getindex(mi::ModelInstance, component::Symbol, name::Symbol)
...
+end
+
+""" returns the values of index i in model m"""
+function getindexvalues(m::Model, i::Symbol)
+    return m.indices_values[i]
+end
+
+""" returns the index labels of the variable or parameter in the given component"""
+function getindexlabels(m::Model, component::Symbol, x::Symbol)
+    metacomp = getmetainfo(m,component)
```
ckingdon95: @davidanthoff @jrising I changed it to just use the word "index". However for variables and parameters originally I called it "getdimensions" because the metacomp.variables and metacomp.parameters attributes are called dimensions. Should that be changed too or is there a reason that they were originally called dimensions instead of indices?

PR #95, opened 2017-02-17
WIP Variables function taking in Model Instance (instead of Model itself)

girishbalaji:
The variables function works using fieldnames but the function does not work on Nullable Instances

### DIFF
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -141,6 +141,15 @@ function variables(m::Model, componentname::Symbol)
     collect(keys(c.variables))
 end
 
+"""
+    variables(mi::ModelInstance, componentname::Symbol)
+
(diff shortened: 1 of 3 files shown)


src/Mimi.jl
```diff
@@ -141,6 +141,23 @@ function variables(m::Model, componentname::Symbol)
...
+
+List all the variables of `componentname` in the ModelInstance 'mi'.
+NOTE: this variables function does NOT take in Nullable instances
+"""
+function variables(mi::Nullable{ModelInstance}, componentname::Symbol)
+    if isnull(mi)
+        error("Cannot get variables when model has not been built yet")
+    end
+    return fieldnames(get(mi).components[componentname].Variables)
+end
```
davidanthoff: Remove this one because it is not an API that users should use.

PR #96, opened 2017-02-17
WIP wrote getdataframe check--still needs test cases

girishbalaji:
checks if variable--component pair exists


davidanthoff:
Add a test for this error path.

PR #116, opened 2017-03-10
[WIP] Plot internal parameter connections

girishbalaji:
Currently using Tom Breloffs plot package.
Components are in blue
Source variables/ target parameters in red


davidanthoff:
The only thing left to do is get rid of that one extra loop. Shows up when showing the ``socioeconomic`` component from FUND.

PR #126, opened 2017-04-10
Update scalar parameters after init

davidanthoff:


jrising:
I think this is solving a different problem, and looking into it, I've realized that our issue was a difference between the latest version and the master branch.  In the master branch, scalars set in the init function do get updated before run_timestep.

Where you have the scalar update, it helps if component B's init relies on variables defined in component A's init.  I think it would be fine to say that that is never supported.  If it is something you want to allow, I think this is fine.

davidanthoff:
> I think this is solving a different problem, and looking into it, I've realized that our issue was a difference between the latest version and the master branch. In the master branch, scalars set in the init function do get updated before run_timestep.

I'm confused. Mimi v0.3.0 was released three weeks ago and the only difference between v0.3.0 and ``master`` is an update to the README at this point.

PR #127, opened 2017-04-14
Expansion of data frame arguments

girishbalaji:
Finished basic expansion of arguments
Should discuss type checking approach
Will add tests


davidanthoff:
This is what we need: 
````julia
 function foo2{T<:Union{Symbol, Tuple}}(x::Pair{Symbol,T})
````

davidanthoff:
- [ ] Write tests
- [ ] Handle case with two 1 dimensional vars that have different indices (throw error)

Issue #129, opened 2017-04-26
`checklabels` uses `parametername` rather than `name`


jrising:
If you use `connectparameter(m::Model, component::Symbol, name::Symbol, parametername::Symbol)` to connect a components parameter named `name` to an external parameter named `parametername`, it will fail if `parametername` is not the same as `name`, because assumes that the component's parameter is named `parametername` rather than `name`.

In fact, `checklabels` takes in both `name` and `parametername`, but does not reference `name` at all.

davidanthoff:
I believe this was fixed, at least on ``master``? I'll close this for now, please reopen if that is wrong.

ckingdon95:
yep it was merged!

jrising:
Thanks!

 - James

PR #133, opened 2017-05-15
checklabels bug

ckingdon95:
issue #129


jrising:
Yes, this seems right to me.  Should we drop `parametername` from this function call?

ckingdon95:
@davidanthoff I think this is ready to be merged

PR #136, opened 2017-05-22
set_external_parameter doc string

ckingdon95:


ckingdon95:
@davidanthoff the travis build had the same amazon error that happened last week (where we need to just re-run it). But I can't find the button you showed me last time to just re-run the build. Does that only show up as an option for you?

PR #140, opened 2017-06-20
David's requested changes

ckingdon95:


ckingdon95:
We definitely need to export Timestep, so that users can define run_timestep(s::comp, ts::Timestep).

We shouldn't have to export TimestepVector/Matrix, but when I remove them, Mimi can't finish loading for some reason, it errors when it tries to load adder.jl saying that TimestepVector is not defined, even though in Mimi.jl, the file TimestepArrays is included before the file for adder. Not sure what the problem is.

Issue #132, opened 2017-05-04
external_parameter_connections stores parameter object, but setparameter creates new objects


jrising:
Currently it is possible for different components to have external parameters of the same name that refer to different numbers.  This is a problem when we want to do a MonteCarlo, because changing the external parameter value (for which-- sub-issue-- you need an `update_external_parameter` function) only changes the value for the most recent component to call `setparameter`.

The mechanism is that component A calls `setparameter`, which creates the ModelParameter object, setting it to be referred to by a given name, and then creates an ExternalParameterConnection using that object (but not the name); then component B comes along, calls `setparameter` with the same name, creates a new ModelParameter object, replaces the reference, but has no effect on the pre-existing ExternalParameterConnections.

Personally, I think that components should be separate namespaces, and that you should never have parameter names that are not connected to components, but at least it should be consistent.

ckingdon95:
I think that maybe in ExternalParameterConnections, the field 'parameter' should be a symbol of the name of the parameter, rather than the actual parameter. The name will reference the model parameter object stored in the dictionary of all external model parameters. How does that sound? @davidanthoff @jrising

jrising:
Yes, that seems like the right way to go.  Thanks!

davidanthoff:
This should be fixed by #142, so I'll close this issue for now. If it does show up again, please reopen!

PR #142, opened 2017-06-22
changes ExternalParameterConnection

ckingdon95:
Now external_parameter is the name of the external_parameter in the model (rather than storing the actual parameter)


ckingdon95:
issue #132

PR #154, opened 2017-09-20
Forces conversion to model type

jrising:
The `convert(Array{T}, x)` call is prone to alias to x in Julia 0.6.  What we really want to do (and here just do explicitly) is allocate a new array of type T to ensure that it is the right type.


davidanthoff:
Thanks! Probably also makes sense to open another PR to merge the same thing onto the ``release-0.3`` branch, so that this will be part of Mimi v0.3.1 as well (the julia 0.6 compatible release).

PR #158, opened 2017-12-05
make dict type {Any, Any} for setleftoverparams

ckingdon95:
the helper function "loadparameters" in Fund was changed and now requires the setleftoverparameters dict type to be {Any, Any}


ckingdon95:
The current version of the master branch on Fund requires:
`function setleftoverparameters(m::Model, parameters::Dict{Any,Any})`

but the version of Fund that's linked in Mimi's Travis build requires
`function setleftoverparameters(m::Model, parameters::Dict{String,Any})`

We could either leave Mimi as it is, or if we want to link the newest version of Fund then we need to update the function signature. Not sure which way we want to go but I wanted to make note that the current master branch of Fund doesn't work with the current master version of Mimi.

PR #159, opened 2017-12-07
fixes error in setindex for nonuniform time indices

ckingdon95:


ckingdon95:
(I noticed this mistake when trying to run page on this version of mimi)

PR #169, opened 2018-01-16
Started documentation of Mimi internals

rjplevin:
David -- this branch off master has a new directory: docs/src/internals, as we discussed.


docs/src/internals/component.md
```diff
@@ -0,0 +1,20 @@
+# Model components
+
+## @defcomp
+The macro ```@defcomp``` defines model components. Some changes are in store:
+
+1. Code simplification using the ```@capture``` macro from MacroTools.jl.
+
+1. Remove the explicit definition of indices via ```regions = Index()``` since regions can be defined on first usage. In any case, indices must be defined (with actual values) at the model level, so ensuring this is the more relevant validity check.
```
davidanthoff: We can merge as is, I'm a little unclear right now whether we want to really do away with it entirely, though. I could imagine something where we keep this, but one passes the default index values in, like ``regions = Index([:US, :WEU])`` or something like that.

docs/src/internals/index.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Renaming of internal structure / types
+We are retaining the distinction between (i) descriptive objects that define the structure (and default values) for models and their components, and (ii) model run-time objects that represent actual model and component instances (i.e., numerous instances of which might be running on a cluster)
+
+* Descriptive object names end with ```Def```, as in ```ModelDef``` and ```ComponentDef```.
+* Run-time objects are the same names without the ```Def```, i.e., ```Model``` and ```Component```.
+
+### Name changes
+* ```Model``` becomes ```ModelDef```
+* ```ModelInstance``` becomes ```Model```
```
davidanthoff: So this is the one case where I'm not yet sure... I could see a version where we have ``ModelDef``, ``ModelInstacne`` and ``Model``. We can discuss today in the meeting.

docs/src/internals/index.md
```diff
@@ -0,0 +1,36 @@
...
+* ```ComponentInstanceInfo``` becomes ```ComponentDef```
```
davidanthoff: I don't think that is right. We need one type for the meta story, and another type for saying that a model holds a certain component.

docs/src/internals/montecarlo.md
```diff
@@ -0,0 +1,136 @@
+# Monte Carlo Simulation support
+
+## @defmcs macro
```
davidanthoff: I think this is nice! I also think we should probably start with the types and functions, and then the macro can be a nice wrapper/shortcut for creating those. But the text here is great.

docs/src/internals/montecarlo.md
```diff
@@ -0,0 +1,136 @@
...
+
+# Save results to the indicated directory
+write(r, dirname)
+```
+
+### Saving ensemble results
+
+Ensemble results will be stored in an object of type ```EnsembleResult```, which is used by all subtypes of ```ModelRunner```. By default, only model outputs tagged for output will be saved in the ```EnsembleResult``` instance. Parameters can be identified as "output" parameters in ```@defmcs```, and/or directly in a concrete subtype of ```ModelRunner```, in which case default values set in ```@defmcs``` can be overridden.
+
+The method ```write(r::ModelRunner, dirname::String)``` will save model results to set of CSV files in the named directory. Initially, we will store the data in [tidy](https://cran.r-project.org/web/packages/tidyr/vignettes/tidy-data.html) format in which variables are in columns and each row represents an observation. This format is verbose but flexible and commonly used by consumers of data such as plotting packages. Other formats can be generated if the need arises.
```
davidanthoff: I think that makes sense, but it is not entirely clear to me how the ``EnsembleResult`` type comes into play here.

docs/src/internals/montecarlo.md
```diff
@@ -0,0 +1,136 @@
...
+  | trialnum | region | year | value |
+  | -------- | ------ | ---- | ----- |
+  | 1 | US | 2010 | 1.1 |
+  | 1 | US | 2015 | 1.9 |
+  | ...|
+  | 1 | CHI | 2010 | 0.2 |
+  | 1 | CHI | 2015 | 0.8 |
+  | ...|
+  
+* Another option for saving ensemble outputs might include writing to any "sink" type that accepts named tuples.
```
davidanthoff: Yes, and there are quite a few!

PR #172, opened 2018-01-30
Explore

davidanthoff:


REQUIRE
```diff
@@ -5,3 +5,4 @@ DataFrames 0.8.4
 Documenter 0.9.0
 NamedArrays 0.5.3
 Plots 0.11.0
+Blink 0.6.0
```
davidanthoff: This should REQUIRE Electron now, not Blink.

deps/build.jl
```diff
@@ -0,0 +1,3 @@
+using Blink
+
+Blink.AtomShell.install()
```
davidanthoff: This whole file can be deleted because Electron.jl handles the necessary binary dependency installation.

src/Mimi.jl
```diff
@@ -10,7 +10,8 @@ export
     connectparameter, setleftoverparameters, getvariable, adder, MarginalModel, ConnectorCompVector,
     ConnectorCompMatrix, getindex, getdataframe, components, variables, getvpd, unitcheck, plot, getindexcount,
     getindexvalues, getindexlabels, delete!, get_unconnected_parameters, Timestep, isfirsttimestep,
-    isfinaltimestep, TimestepVector, TimestepMatrix, hasvalue, update_external_parameter
+    isfinaltimestep, TimestepVector, TimestepMatrix, hasvalue, update_external_parameter,
+    getdataframe_for_parameter, parameters, explore # added for explore.jl
```
davidanthoff: We probably shouldn't export ``getdataframe_for_parameter`` here, right? That seems more like an internal function. Same for ``parameters``, or do we want to make that part of the official API?

Also remove the comment here about ``explore.jl``.

src/explorer/assets/main.html
```diff
@@ -0,0 +1,22 @@
+<!-- Mimi UI -->
+
+<html>
+<header>
+    
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/vega/3.0.8/vega.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/vega-lite/2.0.3/vega-lite.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/vega-embed/3.0.0-rc7/vega-embed.js"></script>
```
davidanthoff: This is fine for the first PR, but @lrennels can you open an issue that in a follow up PR we want to actually just include these files in the repo here and use the local versions, so that things work when one is offline.

src/explorer/assets/main.html
```diff
@@ -0,0 +1,22 @@
...
+<body>
+
+    <div id="variablelist" class="tab"></div>
+    <div id="vis" class="graphic"></div>
+
+    <script src="refresh.js"></script>
+
+</body>
+
+</html>
```
davidanthoff: Very minor nitpick: can you always include one line break at the end of each file?

src/explorer/buildspecs.jl
```diff
@@ -0,0 +1,197 @@
+## Mimi UI
+using Mimi
+using DataFrames
+using JSON
```
davidanthoff: These should all be moved to the toplevel ``Mimi.jl`` file and not appear here.

src/explorer/buildspecs.jl
```diff
@@ -0,0 +1,197 @@
+## Mimi UI
+using Mimi
+using DataFrames
+using JSON 
+
+function getspeclist(model::Mimi.Model)
```
davidanthoff: No need to write ``Mimi.Model``, ``Model`` should be fine (because this function is defined in the same module).

src/explorer/explore.jl
```diff
@@ -0,0 +1,34 @@
+## Mimi UI
+using DataFrames
+using JSON
+using Electron
```
davidanthoff: Move to ``Mimi.jl``.

src/Mimi.jl
lrennels: @davidanthoff  above we do export getdataframe and components, so I figured we'd do the same with the parameter versions, but definitely something we might want to discuss?

rjplevin: getdataframe_for_parameter is going away. I've already modified getdataframe to handle both cases (in the dot-overloading branch). So better to not export it.

lrennels:
@davidanthoff this is ready for code review, the only enhancements to add are multi-line graph color scheme and window size/title

src/Mimi.jl
```diff
@@ -4,13 +4,16 @@ using DataStructures
 using DataFrames
 using Distributions
 using NamedArrays
+using JSON
```
davidanthoff: If we start to use a package here, we also need to list it in REQUIRE, otherwise they will yell at us when we try to tag the package in METADATA.jl. Just add a line with the name to the REQUIRE file.

src/explorer/assets/refresh.js
```diff
@@ -0,0 +1,25 @@
...
+        newButton.setAttribute("class", "tab");
+        
+        // Return a closure with a copy of the spec that's private to the function
+        newButton.onclick = (function() {
+            var spec = speclist[i]["VLspec"];
+
+            return function() {
+                vegaEmbed("#vis", spec, {actions: false});
+            }
+        }())
```
davidanthoff: This seems unnecessary complicated. Can't this be just
````javascript
var spec = speclist[i]["VLspec"];
newButton.onclick = function() {
    vegaEmbed("#vis", spec, {actions: false});
}

File: src/explorer/assets/refresh.js [hunk 11]
```diff
@@ -0,0 +1,25 @@
...
+
+        newButton.appendChild(document.createTextNode(speclist[i]["name"]));
+        document.getElementById("variablelist").appendChild(newButton);
```
davidanthoff: You could cache the return value of ``document.getElementById("variablelist")`` outside of the loop. Probably makes very little difference, but who knows, for very large model results it might help.

src/explorer/buildspecs.jl
```diff
@@ -0,0 +1,195 @@
...
+    #get all components of model
+    comps = Mimi.components(model)
+    for c in comps
+
+        #get all variables of component
+        vars = variables(model, c)
+        for v in vars
+
+            #pull information 
+            name = string("$c : $v") #returns the name of the pair as "component:variable"
```
davidanthoff: No need for the ``string`` here, ``name = "$c : $v"`` should work?

src/explorer/buildspecs.jl
```diff
@@ -0,0 +1,195 @@
...
+
+            #catch errors 
+            try
+                df = getdataframe(model, c, v) #returns the  corresponding dataframe
+
+                #choose type of plot
+                #single value
+                if length(df[1]) == 1
+                    value = df[1][1]
+                    name = string("$c : $v = $value")
```
davidanthoff: Same here, no need to call the ``string`` function here.

src/explorer/buildspecs.jl
```diff
@@ -0,0 +1,195 @@
...
+                    else
+                        spec = createspec_barplot(name, df, dffields)
+                    end
+                end
+
+                #add to spec list
+                push!(allspecs, spec)
+            catch
+                println("could not convert ", name, " to dataframe, skipping ...")
+                continue
```
davidanthoff: I don't think we need the ``continue`` here.

src/explorer/buildspecs.jl
```diff
@@ -0,0 +1,195 @@
...
+            end
+        end
+
+        #get all parameters of component
+        params = parameters(model, c)
+        for p in params
+
+            #pull information 
+            name = string("$c : $p") #returns the name of the pair as "component:parameter"
```
davidanthoff: Same here, no need to call the ``string`` function here.

src/explorer/buildspecs.jl
```diff
@@ -0,0 +1,195 @@
...
+            df = getdataframe_for_parameter(model, c, p) #returns the  corresponding dataframe
+            
+            if !isa(df, Number) #check if there any parameters in this component
+                dffields = names(df)
+                
+                #choose type of plot
+                #single value
+                if length(df[1]) == 1
+                    value = df[1][1]
+                    name = string("$c : $v = $value")
```
davidanthoff: Same here, no need to call the ``string`` function here.

src/explorer/buildspecs.jl
```diff
@@ -0,0 +1,195 @@
...
+            push!(datapart, rowdata)
+        end 
+    else
+        for row in eachrow(df)
+            rowdata = Dict(string(dffields[1])=> row[1], string(dffields[2]) => row[2])
+            push!(datapart, rowdata)
+        end 
+    end
+    return datapart
+end
```
davidanthoff: This function is fine for now, but I suspect it is actually causing some of the performance issues we have. @lrennels can you open a new issue that we should revisit this eventually?

src/explorer/explore.jl
```diff
@@ -0,0 +1,32 @@
+## Mimi UI
+
+global app
```
davidanthoff: Make this ``global app = nothing``

src/explorer/explore.jl
```diff
@@ -0,0 +1,32 @@
...
+include("getparameters.jl")
+
+function explore(model)
+    
+    #get variable data
+    speclist = getspeclist(model)
+    speclistJSON = JSON.json(speclist)
+
+    #start Electron app
+    if !isdefined(:app)
```
davidanthoff: Check here for ``app===nothing``.

src/explorer/getparameters.jl
```diff
@@ -0,0 +1,96 @@
...
+            bottom = fill(NaN, (dim1-final, dim2))
+            data = vcat(top, data, bottom)
+        end
+        df[name] = cat(1,[vec(data[i,:]) for i=1:dim1]...)
+
+        return df
+    else
+        error("Not yet implemented")
+    end
+end
```
davidanthoff: I guess this code should really be somewhere else, but lets leave it for now and then @rjplevin can move it once he integrates things with the dot-overloading branch.

src/explorer/test/01-one-region-model/one-region-model.jl
```diff
@@ -0,0 +1,65 @@
...
+setparameter(my_model, :grosseconomy, :share, 0.3)
+
+#Set parameters for the emissions component
+setparameter(my_model, :emissions, :sigma, [(1. - 0.05)^t *0.58 for t in 1:20])
+connectparameter(my_model, :emissions, :YGROSS, :grosseconomy, :YGROSS)  #Note that connectparameter was used here.
+
+run(my_model)
+
+#Check model results
+my_model[:emissions, :E]
```
davidanthoff: Can this whole file be deleted? We shouldn't have test files in the ``/src`` folder, and this is already in examples, right?

src/explorer/test/test.jl
```diff
@@ -0,0 +1,6 @@
+##  Mimi UI
+using Mimi
+
+## Test 1:  One Region Model
+include(joinpath(@__DIR__, "01-one-region-model/one-region-model.jl"))
+explore(my_model)
```
davidanthoff: Same comment as above, it strikes me that we should delete everything (or move somewhere else) that is in ``src/explorer/test``.

src/explorer/assets/refresh.js
lrennels: @rjplevin this was the result of the debugging work we did to make sure each graph rendered on its own and did not write over the previous ones ... would this simplification work?

src/explorer/buildspecs.jl
rjplevin: All it does is create a closure with the relevant item from the spec list, and assign that function to the button. How would this create a performance problem?

The alternative approach is to assign the spec list to a global, but that prevents multiple windows from working properly. The closure approach embeds the state in the JS in the window.

src/explorer/assets/refresh.js
rjplevin: We tried the approach you suggested, but the value of spec changes between calls, so all the functions end up referencing the same (final) spec. Using a closure solved this problem.

lrennels: Thanks, that's what I thought, so we'll leave it for now.

davidanthoff: Hm, really strange, but yes, lets just leave as is if it works :)

src/explorer/buildspecs.jl
davidanthoff: I think the ``eachrow`` approach is just fairly slow, and indexing into a row with e.g. ``row[1]`` is type unstable, so we are probably leaving a lot of performance on the table here. But we should just profile it and see whether this is where time is spent. If so, there are lots of ways we could make this specific function much faster.

PR #201, opened 2018-03-08
Update timestep to run_timestep

lrennels:
@davidanthoff I updated this function so the one region model tutorial will run properly


rjplevin:
This has been handled in dot-overloading, so I've closed the issue.

PR #208, opened 2018-04-14
Tests

rjplevin:
Converted most of the remaining tests to new API. Very useful: this uncovered a variety of bugs that running several of the models did not. I need to return to the Timestep stuff; the two test files using this fail, but everything else works.


src/mcs/EmpiricalDistribution.jl
```diff
@@ -1,3 +1,7 @@
+try
+    using ExcelReaders
+end
```
davidanthoff: There is a way to handle all of the file loading here without taking a dependency on either anything Excel or CSV, we can discuss in our meeting.

test/runtests.jl
```diff
@@ -1,30 +1,68 @@
 using Mimi
 using Base.Test
 
+# For now, use the warn function; for 0.7/1.0, remove this and use real logging...
+macro info(msg)
+    msg = "\n$msg"
+    :(Base.println_with_color(:light_blue, $msg, bold=true))
+end
```
davidanthoff: Why not just use the ``info`` function? Why a macro?

PR #212, opened 2018-04-22
Extends getdataframe to >2 dims

jrising:
Adds support for variables/parameters of greater than 2 dimensions in `getdataframe`.

I'm not sure if this is applicable to `master`, since it depends on whether `_load_dataframe` or `_load_dataframe_NEW` ends up going into the new version, but since I needed it for myself anyway, I thought I'd share it.

### DIFF
--- a/src/mimi-core.jl
+++ b/src/mimi-core.jl
@@ -500,6 +500,7 @@ function getdataframe(m::Model, mi::ModelInstance, componentname::Symbol, name::
     meta_component_name = Symbol(supertype(typeof(mi.components[componentname])).name.name)
 
     vardiminfo = getdiminfoforvar((meta_module_name,meta_component_name), name)
+
     if length(vardiminfo)==0
         return mi[componentname, name]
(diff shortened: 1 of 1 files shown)


rjplevin:
Thanks, James. Given that v0.4 is approaching obsolescence, I don't think we need to merge this, but I'll leave that decision to David.

The API has changed considerably in master, so I started porting these changes to the new framework, but realized that we'll also need to extend TimestepMatrix to > 2 dimensions to complete this. I'll add an issue for this so we can track it.

rjplevin:
I've implemented this in what will be Mimi v0.5.0. Note that you cannot currently have multiple dimensions with the same name. That is, variable `threeD = Variable(index=[time, region, region])` does not work properly since `getdataframe` uses the name of the dimension (here, `region`) as a column in the resulting dataframe. It does work if you declare a second index, say, `region2` with the same values as `region` and then declare the variable `threeD = Variable(index=[time, region, region2])`. Not optimal, but that's where I've left it for now.

PR #218, opened 2018-05-08
Performance improvements, revisions to tests, and more.

rjplevin:
Sorry for another big PR. Almost all tests pass except for a couple of known problems, e.g., the use of `t::Timestep` as final arg to `run_timestep` remains broken, but barely...

- Reinstated include of plotting.jl since Plots.jl (once again) doesn't break pre-compile
- Created TimestepArray and simplified many functions that were duplicated for TimestepVector/TimestepMatrix, or had to branch on type. Now supports > 2 dimensions.
- Eliminated AbstractTimestepMatrix
- TimestepVector/TimestepMatrix now defined as consts that parameterize TimestepArray with N=1 or N=2, respectively.
- Reimplemented build function to instantiate all vars first so the storage could be referenced directly by connected params.
- Eliminated use of Ref types
- Revised @defcomp to generate functions named run_timestep_MODULENAME_COMPNAME, eliminating the two Val args.
- Revised @defcomp to take optional type arg on final (timestep) parameter, which defaults to ::Int. The only other option (almost) supported is ::Timestep.
- Revised build to cache references to each component's run_timestep function
- Revised call to run_timestep to use the correct type of time argument

### DIFF
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -37,6 +37,7 @@ export
     getproperty,
     get_parameter_value,
     get_variable_value,
+    hasvalue,
     interpolate,
     load_comps,
     modeldef,
@@ -56,7 +57,7 @@ export
 
 include("core/types.jl")
 
-# After loading types and macros, the rest can just be alphabetical
+# After loading types, the rest can just be alphabetical
 include("core/build.jl")
 include("core/connections.jl")
 include("core/defs.jl")
@@ -69,7 +70,7 @@ include("core/model.jl")
 include("explorer/explore.jl")
 include("mcs/mcs.jl")
 include("utils/graph.jl")
-# include("utils/plotting.jl")
+include("utils/plotting.jl")
 include("utils/getdataframe.jl")
 include("utils/lint_helper.jl")
 include("utils/misc.jl")
--- a/src/core/defcomp.jl
+++ b/src/core/defcomp.jl
@@ -46,29 +46,37 @@ function _replace_dots(ex)
 end
 
 function _generate_run_func(module_name, comp_name, args, body)
-
     if length(args) != 4
         error("Can't generate run_timestep; requires 4 arguments but got $args")
+    end
+
+    (p, v, d, t) = args
+
+    if @capture(t, tname_::ttype_)
+        if ttype == :Timestep
+            ttype = :(Mimi.Timestep)
+        end
+        t = tname
+    else
+        ttype = :Int
     end
 
     # replace each expression with its dot-replaced equivalent
     body = [MacroTools.prewalk(_replace_dots, expr) for expr in body]
 
-    # add types to the parameters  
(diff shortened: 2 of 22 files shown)


src/Mimi.jl
```diff
@@ -37,6 +37,7 @@ export
     getproperty,
     get_parameter_value,
     get_variable_value,
+    hasvalue,
```
davidanthoff: I wonder whether we should export this?

src/core/defcomp.jl
```diff
@@ -46,29 +46,37 @@ function _replace_dots(ex)
...
+    end
+
     # replace each expression with its dot-replaced equivalent
     body = [MacroTools.prewalk(_replace_dots, expr) for expr in body]
 
-    # add types to the parameters  
-    (p, v, d, t) = args
+    # Generate unique function name for each component so we can store a function pointer.
+    # (At least in julia 0.6, methods cannot be invoked directly.)
+    funcname = Symbol("run_timestep_$(module_name)_$(comp_name)")
```
davidanthoff: We could probably use an anonymous function here? That way we wouldn't "pollute" the namespace? Can be done later, though.

src/utils/misc.jl
```diff
@@ -23,40 +23,15 @@ function interpolate(values::Vector{T}, ts::Int=10) where T <: Union{Float64, In
...
-#         word_caps = "$(uppercase(word[1]))$(word[2:length(word)])"
-#         to_ret = "$(to_ret)$(word_caps) "
-#     end
-
-#     # Return our string, minus the trailing space that was added
-#     return to_ret[1:length(to_ret) - 1]
-# end
+# MacroTools has a "prettify", so we have to import to "extend"
+# even though our function is unrelated. This seems unfortunate.
+import MacroTools.prettify
```
davidanthoff: I don't understand this?

src/Mimi.jl
rjplevin: It isn't clear to me whether `hasvalue` is ever needed in `run_timestep` functions. It's currently used only in the connector components.

src/core/defcomp.jl
rjplevin: The advantage I see of leaving it this way is that one can type `run_timestep` and hit `TAB` in the REPL and see all the functions, which is potentially useful for developers. This form of pollution has no deleterious effects I'm aware of...

src/utils/misc.jl
rjplevin: A function of the same name is exported by MacroTools, so we can either add another method of the same name, rename, reference ours explicitly (and not export it).
```
julia> using MacroTools

julia> prettify( [TAB]

prettify(ex; lines) in MacroTools at /Users/rjp/.julia/v0.6/MacroTools/src/utils.jl:361
julia> prettify(
```

PR #223, opened 2018-05-17
Eliminated direct use of ExcelReaders in favor of IterableTables

rjplevin:
Implemented XLS/XLSX reading via FileIO and ExcelFiles in mcs/EmpiricalDistribution

### DIFF
--- a/src/mcs/EmpiricalDistribution.jl
+++ b/src/mcs/EmpiricalDistribution.jl
@@ -1,6 +1,8 @@
-try
-    using ExcelReaders
-end
+using FileIO
+using ExcelFiles
+using IterableTables
+using IteratorInterfaceExtensions
+using DataFrames
 
 struct EmpiricalDistribution{T}
     values::Vector{T}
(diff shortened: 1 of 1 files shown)


davidanthoff:
I had thought we would move the whole file loading code out of this, so that one would call this like this:
````julia
d = EmpiricalDistribution(load("foo.xlsx", otherpars=...),:valcolname,:probcolname)
````
The benefit of that would be that we can pass any table in as the first argument, i.e. everything on [this](https://github.com/davidanthoff/IterableTables.jl) list would work. Plus we would not have to take a dependency on FileIO, ExcelFiles etc.

**Comment by rjplevin** [comment]
Sounds good to me.

**Comment by rjplevin** [comment]
What is the (abstract) type returned by `load()`? That is, what should be the type of the first arg to the `EmpiricalDistribution` constructor?

**Comment by davidanthoff** [comment]
There is no common super type, i.e. one can't dispatch on this. You'll just need to have an untyped argument.

I thought a bit more about this over lunch, and another option would be to just expect an iterator of say tuple, or a pair, or a named tuple with two columns or something, i.e. also take out the code where you specify column names. That would make things even more generic, but maybe at the cost of being too generic... In a case where the table one is feeding in doesn't have the correct two columns, one would do something like ``EmpiricalDistribution(load('foo.xlsx") |> @map({_.col1,_.col2}))``, i.e. use Query.jl to transform the source data into the form that is expected by ``EmpiricalDistribution``. In fact, one could write this as ``load('foo.xlsx") |> @map({_.col1,_.col2}) |> EmpiricalDistribution`` in that case.

Maybe we should think hard what use cases this should serve other than the climate sensitivity for the replication exercise.

**Comment by rjplevin** [comment]
Well, there's already the inner constructor `EmpiricalDistribution(values, probs)` for when you have the data available. I was trying to provide a convenient function for loading the data from a file. Maybe we move the latter function to our "extras" folder where we don't worry about dependencies?

**Comment by davidanthoff** [comment]
Ah, I didn't see the inner constructor! That expects two arrays, right?

**Comment by rjplevin** [comment]
Yes. I've now written this, in Mimi/wip (which I'd rather name "extras", personally...)
```
using FileIO
using ExcelReaders
using Mimi: EmpiricalDistribution

function load_vector(path, range, header=false)
    tups = collect(load(path, range, header=header))
    name = fieldnames(tups[1])[1]   # field name of first item in NamedTuple
    map(obj -> getfield(obj, name), tups)
end

function load_empirical_dist(excel_file::AbstractString, 
                             values_range::AbstractString, 
                             probs_range::AbstractString="")                               
    values = load_vector(excel_file, values_range)
    probs = probs_range == "" ? nothing : load_vector(excel_file, probs_range)
    return EmpiricalDistribution(values, probs)
end
```

rjplevin:
EmpiricalDistribution no longer does any loading. See Mimi/WIP/load_empirical_dist.jl for helper functions that do the loading using ExcelReaders.

PR #225, opened 2018-05-21
WIP Tests Travis

lrennels:
This PR will be used to get the Travis tests passing so that the reports will now tell us whether new PRs break something that was previously working.  The testing suite is still being expanded and edited.

### DIFF
--- a/test/test_dependencies.jl
+++ b/test/test_dependencies.jl
@@ -11,8 +11,8 @@ end
 
 #list of URLs of branches of packages to test
 dependencies = [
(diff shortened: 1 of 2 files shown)


src/Mimi.jl
```diff
@@ -71,7 +71,7 @@ include("core/model.jl")
 include("explorer/explore.jl")
 include("mcs/mcs.jl")
 include("utils/graph.jl")
-include("utils/plotting.jl")
+# include("utils/plotting.jl")
```
davidanthoff: Just delete lines like that, there really is no need to keep stuff around in comments. We can always get old code back by looking at the git history.

PR #226, opened 2018-05-21
Mods to MCS subsystem for recent Mimi updates

rjplevin:
- Updated MCS to work with recent Mimi changes
- Removed file loading from EmpiricalDistribution
- Added wip/load_empirical_dist.jl with convenience funcs for use with EmpiricalDistribution
- Added generic "outer loop" processing for iteration over static values outside an MCS


src/core/defs.jl
```diff
@@ -512,15 +512,14 @@ to restore values between trials.
...
-        external_params[key] = obj isa ScalarModelParameter ? ScalarModelParameter(obj.value) : ArrayModelParameter(copy(obj.values), obj.dimensions)
-    end
-
+    external_params = Dict{Symbol, ModelParameter}(key => copy(obj) for (key, obj) in md.external_params)
     return external_params
 end
 
+Base.copy(obj::ScalarModelParameter{T}) where T = ScalarModelParameter{T}(copy(obj.value))
+
+Base.copy(obj::ArrayModelParameter{T})  where T = ArrayModelParameter{T}(copy(obj.values), obj.dimensions)
```
davidanthoff: Can't we use ``deepcopy`` for that and not add a method here? It seems to me that you are redefining ``copy`` here to do what ``deepcopy`` already does.

src/mcs/EmpiricalDistribution.jl
```diff
@@ -33,58 +29,36 @@ struct EmpiricalDistribution{T}
...
-    end
-    return EmpiricalDistribution(values, probs)
-end
-
-# If a column is not identified, use the first column.
-function EmpiricalDistribution(filename::AbstractString)
-    return EmpiricalDistribution(filename, 1)
-end
+#     return EmpiricalDistribution(values, probs)
+# end
```
davidanthoff: Why is this part of the PR? Things like this should really go in their own PR.

src/mcs/mcs_types.jl
```diff
@@ -31,7 +31,14 @@ struct TransformSpec
...
 
-const CorrelationSpec = Tuple{Symbol, Symbol, Float64}
+"""
+Defines a target rank correlation to establish between the two named random vars.
+"""
+struct CorrelationSpec
+    name1::Symbol
+    name2::Symbol
+    value::Float64
+end
```
davidanthoff: I assume this is all done, but in general we could add support for this later: we don't have any model at the moment that needs this.

src/mcs/montecarlo.jl
```diff
@@ -170,58 +188,98 @@ function clear_results(mcs::MonteCarloSimulation)
...
+        clear_results(mcs)
+        
+        # Save the params to be perturbed so we can reset them after each trial
+        original_values = _copy_mcs_params(mcs, md)
 
-        store_trial_results(m, mcs, trialnum)
-    end
+        # Compute how often to print a "Running trial ..." message.
+        count = length(trials)
+        divisor = (count < 50 ? 1 : (count < 500 ? 10 : 100))
```
davidanthoff: We could just use https://github.com/timholy/ProgressMeter.jl here, that would probably result in nicer output.

src/mcs/test/test_defmcs.jl
```diff
@@ -39,11 +45,44 @@ function print_result(m::Model, mcs::MonteCarloSimulation, trialnum::Int)
...
+    println("scen:$scen, rate:$rate")
+end
+
+# Pass as a tuple of pairs
+run_mcs(m, mcs, 10;
+        output_dir="/tmp/Mimi",
+        loop_func=my_loop_func, 
+        loop_args=[:scen => [:low, :med, :high],
+                   :rate => [0.015, 0.03, 0.05]])
+
```
davidanthoff: All the test files seem to be in the wrong folder? None of them should be in the ``/src`` folder, they should all go in the test folder (or subfolders of that).

src/core/defs.jl
rjplevin: There's no need to make a copy of the the dimensions. It's a tradeoff between defining two one line functions and copying stuff we don't need to copy by relying on deepcopy. I could go either way.

src/mcs/EmpiricalDistribution.jl
rjplevin: It's true that this is separate from the MCS stuff. It just gets tedious stashing / branching / unstashing for every small fix...

src/mcs/mcs_types.jl
rjplevin: This is indeed done; I ported my Python code, which already included this. I just changed a `const` declaration of a 3-tuple to a proper struct.

src/mcs/montecarlo.jl
rjplevin: I created an issue for that.

src/mcs/test/test_defmcs.jl
rjplevin: Yeah, these aren't "proper" tests yet. I didn't want to move them to the test folder until I'd integrated them with the testing framework. I can do that next.

src/mcs/EmpiricalDistribution.jl
rjplevin: On the other hand, it's all withing the MCS directory and felt "close enough" to include it. :~)

PR #233, opened 2018-05-30
explorer testing

lrennels:


lrennels:
@davidanthoff I've added some tests for the explorer code, and made the following minor edits to explorer in the meantime to support testing.

- instead of printing a line when a certain spec fails, a `warning` is thrown
- `explore(model, title = "Electron Window")` returns the window 

I'll augment these tests over time, but this should be a good start.

.travis.yml
```diff
@@ -13,11 +13,12 @@ branches:
...
-  - julia -e 'Pkg.build("Mimi")'
-  - julia --check-bounds=yes -e 'Pkg.test("Mimi", coverage=true)'
-  - julia -e 'ENV["PYTHON"]=""; Pkg.add("ExcelReaders")'
-  - julia -e 'include("test/test_dependencies.jl"); run_dependency_tests()'
+  - if [[ `uname` = "Linux" ]]; then TESTCMD="xvfb-run julia"; else TESTCMD="julia"; fi 
+  - $TESTCMD -e'Pkg.clone(pwd())'
+  - $TESTCMD -e 'Pkg.build("Mimi")'
+  - $TESTCMD --check-bounds=yes -e 'Pkg.test("Mimi", coverage=true)'
+  - $TESTCMD -e 'ENV["PYTHON"]=""; Pkg.add("ExcelReaders")'
+  - $TESTCMD -e 'include("test/test_dependencies.jl"); run_dependency_tests()'
```
davidanthoff: I think we only need to use ``$TESTCMD`` for this line, all the other lines should just work with the plain ``julia`` command, i.e. don't need to be changed.

lrennels: I actually tried that and the Linux build was failing, I'll remove them now and push to double heck.  I think they may be at least needed for lines 18 and 19, DataVoyager.jl has `$TESTCMD -e 'Pkg.clone(pwd()); Pkg.build("DataVoyager"); Pkg.test("DataVoyager"; coverage=true)'`

lrennels: @davidanthoff sorry which line number did you think we needed `$TESTCMD` for?

davidanthoff: Ah, sorry, I kind of messed it up. I think it is needed for only line 19.

PR #235, opened 2018-06-06
Variable Timestep

lrennels:


src/core/defs.jl
```diff
@@ -132,17 +132,29 @@ description(def::DatumDef) = def.description
...
+
+function years_array(md::ModelDef)
+    keys::Vector{Int} = dim_keys(md, :time)
+    return keys
 end
 
-function step_size(values::Vector{Int})
-    # N.B. assumes that all timesteps of the model are the same length
-     return length(values) > 1 ? values[2] - values[1] : 1
+function years_array(values::Vector{Int})
```
rjplevin: What's the purpose if `years_array(values) = values` ?

src/core/build.jl
```diff
@@ -12,9 +12,20 @@ function _instance_datatype(md::ModelDef, def::DatumDef, start::Int)
...
-        step = step_size(md)
-        T = TimestepArray{dtype, num_dims, start, step}
+    else   
+
+        #T = TimestepArray{AbstractTimestep, dtype, num_dims}
+
+        # LFR-TBD:  There may be a more elegant way to carry out the logic below.  
+        # We need to access years and stepsize, so this might have to do with
+        # how the isuniform function is used etc.
+        years = years_array(md)
```
rjplevin: One small improvement would be to test for `isuniform` when the time dimension is set, and store this in the `ModelDef`. Then you could retrieve the years array only when its needed (i.e., for variable timesteps). Same pattern in connections.jl.

src/core/defs.jl
```diff
@@ -191,26 +203,38 @@ function set_dimension!(md::ModelDef, name::Symbol, keys::Union{Int, Vector, Tup
...
-# helper function for setindex; used to determine if the provided time values are a uniform range.
+# LFR-TBD:  We might be able to simplify this function, and we should think through
+# it's return carefully as it is at the root of several functions.  Also, do we
+# want to be able to call this on a model too (like years_array?) ... otherwise
+# we often call years_array and then isuniform.
+#
+# helper function for setindex; used to determine if the provided time values are 
+# a uniform range. The function returns -1 if the vector is not uniform, 
+#otherwise it returns the timestep length aka stepsize
 function isuniform(values::Vector)
```
rjplevin: You can simplify this to
```
all_equal(values) = all(map(val -> val == values[1], values[2:end]))

# I'd keep this generic since it's useful, and for clarity, is* functions should return booleans
isuniform(values) = all_equal(diff(values))

# A function to check length(values), etc., and return -1 if variable or the step length if fixed.
function get_timestep(values) 
   ... 
end
```

src/core/instances.jl
```diff
@@ -292,10 +300,20 @@ function Base.run(mi::ModelInstance, ntimesteps::Int=typemax(Int),
...
+    step  = isuniform(t)
+
+    if step != -1
+        comp_clocks = [Clock{Timestep}(start, step, stop) for (start, stop) in zip(starts, stops)]
+    else
+        comp_clocks = Array{Clock{VariableTimestep}}(length(starts))
+        for i = 1:length(starts)
+            start_index = findfirst(t, starts[i])
+            stop_index = findfirst(t, stops[i])
+            years = t[start_index:stop+index]
```
rjplevin: `stop+index` looks like a typo for `stop_index`

src/core/defs.jl
lrennels: Good question, I'll take a look.  I was mainly just copying the `step_size` functions so that I can replace uses of `step_size` with `years_array`when needed, but this is probably superfluous.

src/core/build.jl
lrennels: Good idea, I was thinking that we might be able to put isuniform calls somewhere more convenient.  I think this pattern appears once more in `defs.jl`, so that's three times we could make smoother.

lrennels: Taken care of via discussion today, still should take a look at new code to find places to simplify/optimize

src/core/defs.jl
lrennels: Done, with a few modifications to handle edge cases.

src/core/defs.jl
```diff
@@ -368,15 +409,23 @@ end
...
     comp_def = compdef(md, comp_name)
     start = start_period(comp_def)
     stop  = stop_period(comp_def)
-    step  = step_size(md)
-    return Int((stop - start) / step + 1)
+    years = years_array(md)
+    start_index = findfirst(years, start)
+    stop_index = findfirst(years, stop)
+    return Int(size(years[start:stop]))
 end
```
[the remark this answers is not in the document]
lrennels: Done

src/core/instances.jl
lrennels: Yep :)

lrennels:
@davidanthoff @rjplevin this PR is ready to be reviewed.  After it is merged I will do a second small PR to carry out:

src/core/time.jl
```diff
@@ -1,38 +1,66 @@
...
+function is_stop(ts::VariableTimestep{Years}) where {Years}
+	return gettime(ts) == Years[end]
+end
+
+function finished(ts::Timestep{Start, Step, Stop}) where {Start, Step, Stop}
 	return gettime(ts) > Stop
 end
 
-function next_timestep{Start, Step, Stop}(ts::Timestep{Start, Step, Stop})
+function finished(ts::VariableTimestep{Years}) where {Years}
```
rjplevin: Looks like these two funcs can be combined, since the type parameters of the two Timestep variants are not used at all.

```
function next_timestep(ts::Timestep{Start, Step, Stop}) where {Start, Step, Stop}
	if finished(ts)
			error("Cannot get next timestep, this is final timestep.")
	end
	return Timestep{Start, Step, Stop}(ts.t + 1)
end

function next_timestep(ts::VariableTimestep{start_times}) where {start_times}
	if finished(ts)
		error("Cannot get next timestep, this is final timestep.")
	end
	return VariableTimestep{start_times}(ts.t + 1)		
end
```
into
```
function next_timestep(ts::T) where {T :< AbstractTimestep}
	if finished(ts)
	    error("Cannot get next timestep, this is final timestep.")
	end
	return T(ts.t + 1)
end
```

src/core/defs.jl
```diff
@@ -132,17 +132,18 @@ description(def::DatumDef) = def.description
...
-    keys::Vector{Int} = dim_keys(md, :time) # keys are, e.g., the years the model runs
-    return length(keys) > 1 ? keys[2] - keys[1] : 1
+function start_step(values::Vector{Int})
+     return values[1], (length(values) > 1 ? values[2] - values[1] : 1)
 end
 
-function step_size(values::Vector{Int})
-    # N.B. assumes that all timesteps of the model are the same length
-     return length(values) > 1 ? values[2] - values[1] : 1
+function starttimes(md::ModelDef)
```
rjplevin: The function `starttimes` is redundant with `timelabels(md::ModelDef)`, and I think the latter is a clearer name. I note that `timelabels` isn't currently used, and on further reflection, should be renamed `time_labels`.

src/core/defs.jl
```diff
@@ -132,17 +132,18 @@ description(def::DatumDef) = def.description
 
 unit(def::DatumDef) = def.unit
 
-# step_size(md::ModelDef) = step_size(indexvalues(md))
+function start_step(md::ModelDef)
```
rjplevin: I find the term `start` in `start_step` a bit confusing, since it refers both to the "first" timestep "start" (time label) value. I think it would be clearer to rename `start_step` => `first_and_step` since it returns both the first value and the first step?

Also, reduce redundancy by having first of these methods call the second (short funcs get inlined anyway): 
```
function first_and_step(md::ModelDef)
    keys::Vector{Int} = timelabels(md) # keys are the start_times of the model runs
    return first_and_step(keys)
end

function first_and_step(values::Vector{Int})
     return values[1], (length(values) > 1 ? values[2] - values[1] : 1)
end
```

Assuming `Vector{Int}` is fine for this first pass, but thinking about non-Int time values, we could remove the Int assumptions but parameterize the second func for efficiency:
```
function first_and_step(md::ModelDef)
    labels = timelabels(md)       # labels are the start_times of the model runs
    return first_and_step(labels)
end

function first_and_step(values::Vector{T}) where T
     return values[1], (length(values) > 1 ? values[2] - values[1] : T(1))   # I'm not sure this works for time values...
end
```

lrennels:
@rjplevin I addressed your changes and made note of further updates we'll want to make after this is merged.  Ready to go when you approve!

src/core/build.jl
```diff
@@ -12,9 +12,22 @@ function _instance_datatype(md::ModelDef, def::DatumDef, start::Int)
     elseif dims[1] != :time
         T = Array{dtype, num_dims}
     
-    else
-        step = step_size(md)
-        T = TimestepArray{dtype, num_dims, start, step}
+    else   
+        # OLD WAY
+        # step = step_size(md)
+        # T = TimestepArray{dtype, num_dims, start, step}
```
davidanthoff: Lets remove old stuff like that. We can always look at the git history if we want to see how it was done previously, but this kind of stuff tends to accumulate, I think it is better to just delete it.

src/core/build.jl
```diff
@@ -12,9 +12,22 @@ function _instance_datatype(md::ModelDef, def::DatumDef, start::Int)
...
+
+        if isuniform(md)
+            #take the start from the function argument, not the model def
+            ~, stepsize = first_and_step(md)
```
davidanthoff: Don't assign something to ``~``, that symbol is already bound to a function. You can use ``_`` if you don't care about the first unpacked argument (i.e. ``_, stepsize = ...``).

src/core/connections.jl
```diff
@@ -107,11 +107,30 @@ function connect_parameter(md::ModelDef,
...
-        T = eltype(backup)
-
+        T = eltype(backup)        
+        
         dim_count = length(dst_dims)
-        values = dim_count == 0 ? backup : TimestepArray{T, dim_count, start, step}(backup)
+
+        # OLD WAY
+        # step = step_size(md)
+        # values = dim_count == 0 ? backup : TimestepArray{T, dim_count, start, step}(backup)
```
davidanthoff: Remove

src/core/connections.jl
```diff
@@ -107,11 +107,30 @@ function connect_parameter(md::ModelDef,
...
+        
+        if dim_count == 0
+            values = backup
+        else
+            
+            if isuniform(md)
+                #use the start from the comp_def not the ModelDef
+                ~, stepsize = first_and_step(md)
```
davidanthoff: Don't assign to ``~``

src/core/connections.jl
```diff
@@ -193,10 +212,14 @@ function set_leftover_params!(md::ModelDef, parameters::Dict{T, Any}) where T
...
                 if num_dims in (1, 2) && param_dims[1] == :time   # array case
                     value = convert(Array{md.number_type}, value)
-                    # start = indexvalues(md, :time)[1]
-                    start = dim_keys(md, :time)[1]
-                    step = step_size(md)
-                    values = get_timestep_instance(eltype(value), start, step, num_dims, value)
+                    # OLD WAY
+                    # start = dim_keys(md, :time)[1]
+                    # step = step_size(md)
+                    # values = get_timestep_instance(eltype(value), start, step, num_dims, value)
```
davidanthoff: Don't keep old version around here.

src/core/defs.jl
```diff
@@ -132,17 +132,18 @@ description(def::DatumDef) = def.description
 
 unit(def::DatumDef) = def.unit
 
-# step_size(md::ModelDef) = step_size(indexvalues(md))
+function first_and_step(md::ModelDef)
+    keys::Vector{Int} = dim_keys(md, :time) # keys are, e.g., the times of the model runs
```
davidanthoff: I don't think we want to use these kind of type declarations on the left side of ``=``. I'm not even sure what they do, but I don't think they do what we want (namely ensure the type of ``keys``).

src/core/defs.jl
```diff
@@ -132,17 +132,18 @@ description(def::DatumDef) = def.description
...
-    return length(keys) > 1 ? keys[2] - keys[1] : 1
+function first_and_step(values::Vector{Int})
+     return values[1], (length(values) > 1 ? values[2] - values[1] : 1)
 end
 
-function step_size(values::Vector{Int})
-    # N.B. assumes that all timesteps of the model are the same length
-     return length(values) > 1 ? values[2] - values[1] : 1
+function time_labels(md::ModelDef)
+    keys::Vector{Int} = dim_keys(md, :time)
```
davidanthoff: Same here, don't use type declaration on the left side of ``=``.

src/core/defs.jl
```diff
@@ -291,13 +296,33 @@ function set_parameter!(md::ModelDef, comp_name::Symbol, param_name::Symbol, val
...
         if comp_param_dims[1] == :time
             T = eltype(value)
-            start = start_period(comp_def)
-            dur = step_size(md)
 
-            values = num_dims == 0 ? value : TimestepArray{T, num_dims, start, dur}(value)
+            # OLD WAY
+            # start = start_period(comp_def)
+            # dur = step_size(md)
+            # values = num_dims == 0 ? value : TimestepArray{AbstractTimestep, T, num_dims}(value)
```
davidanthoff: Delete old code.

src/core/defs.jl
```diff
@@ -291,13 +296,33 @@ function set_parameter!(md::ModelDef, comp_name::Symbol, param_name::Symbol, val
...
+
+            if num_dims == 0
+                values = value
+            else
+                start = start_period(comp_def)
+
+                if isuniform(md)
+                    #want to use the start from the comp_def not the ModelDef
+                    ~, stepsize = first_and_step(md)
```
davidanthoff: Don't assign to ``~``.

src/core/defs.jl
```diff
@@ -368,15 +393,23 @@ end
 #
 
 # Return the number of timesteps a given component in a model will run for.
+# function getspan(md::ModelDef, comp_name::Symbol)
+#     comp_def = compdef(md, comp_name)
+#     start = start_period(comp_def)
+#     stop  = stop_period(comp_def)
+#     step  = step_size(md)
+#     return Int((stop - start) / step + 1)
+# end
```
davidanthoff: Delete old code

src/core/instances.jl
```diff
@@ -289,13 +297,28 @@ function Base.run(mi::ModelInstance, ntimesteps::Int=typemax(Int),
...
-
+    
     starts = mi.starts
     stops = mi.stops
-    step  = step_size(t)
 
-    comp_clocks = [Clock(start, step, stop) for (start, stop) in zip(starts, stops)]
+    # OLD WAY
+    # step  = step_size(t)
+    # comp_clocks = [Clock(start, step, stop) for (start, stop) in zip(starts, stops)]
```
davidanthoff: Delete old code.

src/core/instances.jl
```diff
@@ -289,13 +297,28 @@ function Base.run(mi::ModelInstance, ntimesteps::Int=typemax(Int),
...
+    
     
+    if isuniform(t)
+        ~, stepsize = first_and_step(t)
```
davidanthoff: Don't assign to ``~``

src/core/time.jl
```diff
@@ -1,38 +1,62 @@
+#  General TODO:  
+
 #
-#  TIMESTEP
+#  1. TIMESTEP
```
davidanthoff: We should keep TODOs in issues, not in the code.

src/core/time.jl
```diff
@@ -72,150 +108,318 @@ const AnyIndex = Union{Int, Vector{Int}, Tuple, Colon, OrdinalRange}
...
+function Base.getindex(mat::TimestepMatrix{VariableTimestep{start_times}, T}, idx1::AnyIndex, idx2::AnyIndex) where {T, start_times}
 	return mat.data[idx1, idx2]
 end
 
-function Base.setindex!(mat::TimestepMatrix{T, Start, Step}, val, ts::Timestep{Start, Step, Stop}, idx::AnyIndex) where {T, Start, Step, Stop}
+# function Base.setindex(mat::TimestepMatrix, val, ts::T, idx::AnyIndex) where {T <: AbstractTimestep}
+# 	setindex!(mat.data, val, ts.t, idx)
+# end
+
+function Base.setindex(mat::TimestepMatrix{Timestep{Start, Step}, T}, val, ts::Timestep{Start, Step, Stop}, idx::AnyIndex) where {T, Start, Step, Stop}
```
davidanthoff: Just had a thought (OH NO): normally we use all caps for type parameters... Wondering whether we should do the same throughout, i.e. ``START``, ``STEP`` and ``STOP`` etc...

Maybe we can open an issue for that, we shouldn't stop this PR from merging for that, but we might want to change it down the road.

src/core/types.jl
```diff
@@ -1,40 +1,67 @@
+# General TODO:
+# 1.  Change the names of timesteps to AbstractTimestep, FixedTimestep, and
+# VariableTimestep (as of now FixedTimestep is just called Timestep) 
+# 2.  Change Start, Stop to First, Last, which are more logical names in
+# context
```
davidanthoff: These TODOs seem to be done? Delete this comment?

PR #244, opened 2018-06-20
mcs2

rjplevin:
This is working but is not 100% complete. I wanted to merge now, though, before it gets too huge.

Currently, trial data are produced in a NamedTuple in all cases. We discussed a lighter weight approach when possible, but I figured we'd see if this simpler approach was adequate. I also need to profile the code a bit more.

I also need to move the test code to the test directory... I'll do that in the next PR.


davidanthoff:
Oh, but one thing: we should probably try to get the CI runs to pass. I think right now it is missing ``ProgressMeter`` in some REQUIRE file?

rjplevin:
Ah, right. Two questions about that:

1. What about IterTools, IterableTables, TableTraits? Which of these needs to be REQUIRE'd?
2. Should we specify the current version of packages we're testing with?

davidanthoff:
If we use them, we need to list them in REQUIRE. I think we can leave the version bound out for now and hope for the best :)

PR #248, opened 2018-06-23
mcs3

rjplevin:
Merged MCS tests into Mimi's test. Also debugged test_timesteps.jl. (Should have been another PR?)

Although I merged master into mcs3 before issuing this PR, Lisa's changes in connections.jl are showing up here. I don't understand why, but hopefully this doesn't cause any problems. (Was her last PR merged yet?)


lrennels:
@rjplevin you may run into test failures for Travis's Linux build having to do with ExcelFiles and Miniconda.  I was running into them earlier this year and kicked the can down the road, but if we are back to needing ExcelFiles then we may need to dig into what's going on there.  We can take a look once the conflicts are resolved and the tests run.

PR #254, opened 2018-07-02
Doc to Port v0.4.0 to 0.5.0

lrennels:
This document is meant to assist in porting Mimi models from v0.4.0 to 0.5.0.  It is structured after the `userguide.md` doc, and outlines the most commonly needed changes, but not all available or possible changes.


wip/integrationguide.md
```diff
@@ -0,0 +1,110 @@
...
+```julia
+@defcomp component1 begin
+
+    # First define the state this component will hold
+    savingsrate = Parameter()
+
+    # Second, define the (optional) init function for the component
+    function init(p, v, d)
+    end
+
```
rjplevin: I think we need to firm up the semantics of `init()` vs `run_timestep()`. 

- Is `init` called _before_ any timesteps, or is it called on the first timestep? 
- If init is called on the first timestep, should `run_timestep` run only from the 2nd timestep forward?
- What if a component is run with an offset (i.e., it starts after the first global timestep)?

Currently, `init()` is called for all components before the timestep loop, which at least is easy to explain, but then `run_timestep()` needs to avoid redundant behavior in 1st real timestep. Perhaps we should start by enumerating the use cases?

wip/integrationguide.md
```diff
@@ -0,0 +1,110 @@
...
+In an effort to standardize the function naming protocol within Mimi, and to streamline it with the standards of Julia, several function names have been changed.  The table below lists a **subset** of these changes, focused on the exported API functions most commonly used in model construction.  
+
+| Old Syntax                | New Syntax                |
+| ------------------------  |:-------------------------:|
+|`connectparameter`         |`connect_parameter`        |
+|`setleftoverparameters`    |`set_leftover_params!`     |
+|`setparameter`             |`set_parameter!`           |
+|`adddimension`             |`add_dimension`            |
+|`setindex`                 |`set_dimension!`           |
```
rjplevin: Should it be `add_dimension!`?

wip/integrationguide.md
```diff
@@ -0,0 +1,110 @@
...
+
+The new version does, however, include a new UI tool that can be used to visualize model results.  This `explore` funcion is described in the User Guide under **Advanced Topics**.
+
+## Advanced Topics
+
+### Timesteps and available functions
+
+As previously mentioned, some relevant function names have changed.  Below is a subset of such changes related to timesteps and available functions.
+
+| Old Syntax                | New Syntax                |
```
rjplevin: Maybe add explanation for the change, i.e., to eliminate ambiguity about whether value given as "stop" was the final period run or the first period _not_ run. New naming is clearer.

wip/integrationguide.md
lrennels: @rjplevin This is a good point.  I'm a novice on the `init()` function, haven't really used it or seen it used much, but we should probably discuss this further to firm it up in the documentation.  That wouldn't quite belong in this doc as much as the general documentation.  I'll make a note that we should discuss further with @davidanthoff.

wip/integrationguide.md
lrennels: @rjplevin no, this is actually currently `add_dimension` ... perhaps just because it adds one but does not set it like `set_dimension!`?  If `add_dimension!` follows convention better I can change that.

davidanthoff:
This is super helpful! I think we should maybe just make it part of the normal documentation?

lrennels:
@rjplevin @davidanthoff I think this is ready to be merged now, we can add to it if we think of other things

wip/integrationguide.md
rjplevin: I was thinking it should have the "!" for consistency, since it modifies the model definition.

lrennels: I'd tend to agree, I can make a small PR to change this

PR #265, opened 2018-07-13
Add methods to set_external_param!

lrennels:
This PR adds methods to `set_external_param!` so that it has functionality to call `set_external_scalar_param!` and `set_external_array_param!`.  This allowed simplification of `set_leftover_params!`. 

Issue:  mimi-page-2009.jl Issue 2: setdistinctparameter()

**Outstanding Issues:** do we need to `decache(m)` after `set_external_param!`

### DIFF
--- a/src/core/connections.jl
+++ b/src/core/connections.jl
@@ -202,7 +202,7 @@ function set_leftover_params!(md::ModelDef, parameters::Dict{T, Any}) where T
             param_dims = parameter_dimensions(md, comp_name, param_name)
             num_dims = length(param_dims)
 
-            set_external_param!(md, param_name, value, param_dims, num_dims)
+            set_external_param!(md, param_name, value; param_dims = param_dims, num_dims = num_dims)
             
             #TODO:  if tests pass and we like this change to set_external_param!, delete the commented out code below
             # if num_dims == 0    # scalar case
@@ -265,8 +265,8 @@ function set_external_param!(md::ModelDef, name::Symbol, value::ModelParameter; 
     md.external_params[name] = value
 end
 
-function set_external_param(md::ModelDef, name::Symbol, value::Number; param_dims = [], num_dims = 0)
-    set_external_scalar_param!(md, param_name, value)
+function set_external_param!(md::ModelDef, name::Symbol, value::Number; param_dims = [], num_dims = 0)
+    set_external_scalar_param!(md, name, value)
 end
 
 function set_external_param!(md::ModelDef, name::Symbol, value::Union{AbstractArray, Range, Tuple}; param_dims = [], num_dims = 0)
@@ -280,7 +280,7 @@ function set_external_param!(md::ModelDef, name::Symbol, value::Union{AbstractAr
          values = value
     end
 
-    set_external_array_param!(md, param_name, values, param_dims)
+    set_external_array_param!(md, name, values, param_dims)
 end
 
 """
--- a/src/core/connections.jl
+++ b/src/core/connections.jl
@@ -199,26 +199,9 @@ function set_leftover_params!(md::ModelDef, parameters::Dict{T, Any}) where T
         # check whether we need to set the external parameter
         if ! haskey(md.external_params, param_name)
             value = parameters[string(param_name)]
-            param_dims = parameter_dimensions(md, comp_name, param_name)
-            num_dims = length(param_dims)
 
-            set_external_param!(md, param_name, value; param_dims = param_dims, num_dims = num_dims)
-            
-            #TODO:  if tests pass and we like this change to set_external_param!, delete the commented out code below
-            # if num_dims == 0    # scalar case
-            #     set_external_scalar_param!(md, param_name, value)
+            set_external_param!(md, param_name, value; comp_name = comp_name)
 
-            # else
-            #     if num_dims in (1, 2) && param_dims[1] == :time   # array case
-            #         value = convert(Array{md.number_type}, value)
-
-            #         values = get_timestep_instance(md, eltype(value), num_dims, value)
-                    
-            #     else
-            #         values = value
-            #     end
-            #     set_external_array_param!(md, param_name, values, param_dims)
-            # end
         end
         connect_parameter(md, comp_name, param_name, param_name)
     end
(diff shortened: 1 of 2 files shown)


src/core/connections.jl
```diff
@@ -258,10 +261,28 @@ function add_external_param_conn(md::ModelDef, conn::ExternalParameterConnection
     push!(md.external_param_conns, conn)
 end
 
-function set_external_param!(md::ModelDef, name::Symbol, value::ModelParameter)
+function set_external_param!(md::ModelDef, name::Symbol, value::ModelParameter; param_dims = [], num_dims = 0)
```
rjplevin: It looks like the two optional parameters can be replaced by one, `comp_name`, allowing `param_dims` and `num_dims` to be computed in the `set_external_param!` rather than by the caller (which ends up propagating two optional parameters through several methods):
```
 param_dims = parameter_dimensions(md, comp_name, param_name)
 num_dims = length(param_dims)
```

src/core/connections.jl
```diff
@@ -258,10 +261,28 @@ function add_external_param_conn(md::ModelDef, conn::ExternalParameterConnection
...
     md.external_params[name] = value
 end
 
+function set_external_param!(md::ModelDef, name::Symbol, value::Number; param_dims = [], num_dims = 0)
```
rjplevin: Doesn't need the optional parameters if computed where needed...

src/core/model.jl
```diff
@@ -62,11 +62,20 @@ function connect_parameter(m::Model, dst::Pair{Symbol, Symbol}, src::Pair{Symbol
     connect_parameter(m.md, dst[1], dst[2], src[1], src[2], backup; ignoreunits=ignoreunits, offset=offset)
 end
 
-function set_external_param!(m::Model, name::Symbol, value::ModelParameter)
+function set_external_param!(m::Model, name::Symbol, value::ModelParameter; param_dims = [], num_dims = 0)
     set_external_param!(m.md, name, value)
```
rjplevin: optional args are not necessary here

src/core/model.jl
```diff
@@ -62,11 +62,20 @@ function connect_parameter(m::Model, dst::Pair{Symbol, Symbol}, src::Pair{Symbol
...
     decache(m)
 end
 
+function set_external_param!(m::Model, name::Symbol, value::Number; param_dims = [], num_dims = 0)
```
rjplevin: Optional args not needed here either.

src/core/model.jl
```diff
@@ -62,11 +62,20 @@ function connect_parameter(m::Model, dst::Pair{Symbol, Symbol}, src::Pair{Symbol
...
+    set_external_param!(m.md, name, value)
+    decache(m)
+end
+
+function set_external_param!(m::Model, name::Symbol, value::Union{AbstractArray, Range, Tuple}; param_dims = [], num_dims = 0)
```
rjplevin: Compute `param_dims` and `num_dims` here rather than passing as args. We need to consider whether setting an external parameter should invalidate the model instance (i.e., if we should call `decache(m)` here. I'll have to review the MCS code but I think we can avoid invalidation when updating a value, but not if we're adding a new external parameter.

src/core/connections.jl
lrennels: @rjplevin great, I wasn't sure how to access the information without other inputs.  So this will be a different way to access the same values as below.  Then as you said we can remove optional parameters from everything ... I was hoping we'd be able to do that!

```
param_dims = Mimi.parameter_dimensions(m, component, name)    
num_dims = length(size(value))
```

src/core/model.jl
lrennels: I was thinking I would calculate it within ...

`set_external_param!(md::ModelDef, name::Symbol, value::Union{AbstractArray, Range, Tuple})`

... since that's the only place we actually end up needing those parameters

src/core/connections.jl
lrennels: @rjplevin the issue here is that we cannot call `param = external_param(md, name)`, because the parameter in question is currently unconnected and thus not in the list of external parameters.  Instead, the only way to access it's dimensions is to look at the component itself, but we don't want to have to pass the component name through if we can avoid it.  **Assuming that all external parameters must be uniquely named** (can we definitely assume this though?), we could go call `unconnected_params(md)` to get the component name, which adds some work but avoids passing around the component name.   Alternatively we can pass in the component name and keep that one extra optional arg.

Previous `set_leftover_params!`:

```
for (comp_name, param_name) in leftovers
        # check whether we need to set the external parameter
        if ! haskey(md.external_params, param_name)
            value = parameters[string(param_name)]
            param_dims = parameter_dimensions(md, comp_name, param_name)
            num_dims = length(param_dims)

            set_external_param!(md, param_name, value; param_dims = param_dims, num_dims = num_dims)
```

rjplevin: I see... I think having the one optional arg (`comp_name`) isn't so bad. If it's not provided, we can look for the param name using `unconnected_params`. (Might write a function for that so we can use it elsewhere.)

rjplevin: I don't that that is enforced. At some point, we might want to revisit the treatment of external vars...

lrennels: Agreed, I think `set_leftover_params()` itself is on our backlog list of things to revisit, but for now this change will simplify code and work.

src/core/connections.jl
```diff
@@ -262,6 +249,26 @@ function set_external_param!(md::ModelDef, name::Symbol, value::ModelParameter)
     md.external_params[name] = value
 end
 
+function set_external_param!(md::ModelDef, name::Symbol, value::Number; param_dims::Array{Symbol} = nothing)
```
rjplevin: For `param_dims::Array{Symbol} = nothing` to work with `nothing` as a default value, you'll need to define the type as `param_dims::Union{Void, Array{Symbol}} = nothing`

src/core/connections.jl
```diff
@@ -262,6 +249,26 @@ function set_external_param!(md::ModelDef, name::Symbol, value::ModelParameter)
...
+    set_external_scalar_param!(md, name, value)
+end
+
+function set_external_param!(md::ModelDef, name::Symbol, value::Union{AbstractArray, Range, Tuple}; param_dims::Array{Symbol} = nothing)
+
```
rjplevin: Ditto on `param_dims = nothing`

src/core/model.jl
```diff
@@ -67,6 +67,15 @@ function set_external_param!(m::Model, name::Symbol, value::ModelParameter)
     decache(m)
 end
 
+function set_external_param!(m::Model, name::Symbol, value::Number; param_dims::Array{Symbol} = nothing)
+    set_external_param!(m.md, name, value; param_dims = param_dims)
```
rjplevin: Ditto.

lrennels:
@rjplevin done!

PR #268, opened 2018-07-19
Fix MCS output directory nesting

ckingdon95:
@rjplevin this was the directory bug I was mentioning. Currently folders for multiple models get nested within one another because 'output_dir' is repeatedly being reused


ckingdon95:
@davidanthoff I think I don't have write access, can you add me somehow?

davidanthoff:
Done!

Issue #247, opened 2018-06-23
Conceptual question about Timesteps


rjplevin:
I spent a couple of hours debugging test_timesteps.jl today. I found a conceptual error that could be resolved various ways. I'm looking for input from the experienced Mimi modelers.

Note that we decided to eliminate passing time to `run_timestep` as an integer index, instead passing either a `FixedTimestep` or `VariableTimestep`. So this problem may disappear in that transition.

The function `addcomponent` has an optional argument `first` that indicates the first time (a year, for now) that this component should run. A component with an "offset" like this can be connected to a component with a different offset, or no offset. This scrambles the interpretation of `t` in `run_timestep`, since the same index implies different years in the 2 components. This allows "shorter" components to allocate less storage than they would otherwise have to.

This should be resolved when moving exclusively to `ts::AbstractTimestep` in `run_timestep` functions, since we're explicit about year and implicit about array position, rather than the other way around.

It's not clear to me whether this "offset" feature is actually used, and if so, how this was working in version 0.4.0. I'd like to be sure the new version doesn't violate any assumptions in existing models.

davidanthoff:
I think our idea had been that this whole feature would only work if one uses ``Timestep`` indices, not with just ``Int`` indices. I don't think it is being used anywhere right now. I think the best way forward on that is to just ditch the whole ``Int`` timestep variant, and require folks to use the proper timestep types.

lrennels:
This makes sense to me, we currently do have portions of the code in Mimi that are meant to support this old way of indexing.  I could remove these before the release if that is desirable.

lrennels:
Notes for removing `Int` tilmestep variant:

- remove `ci.use_integer_time` flag

```
mutable struct ComponentInstance{TV <: ComponentInstanceVariables, TP <: ComponentInstanceParameters}
    ...
    use_integer_time::Bool                  # will be deprecated when switching to timesteps in all cases
```

- remove methods in `time.jl` with comment # int indexing version supports old style components

rjplevin:
For clarity, the change should be to this, with the `where` bit after the parenthesized arguments. (Inside doesn't create the correct code...)

```
    func = :(
        function $(funcname)($(p)::Mimi.ComponentInstanceParameters, $(v)::Mimi.ComponentInstanceVariables, 
                             $(d)::Dict{Symbol, Vector{Int}}, $(t)::T) where {T <: Mimi.AbstractTimestep}
            $(body...)
            
            return nothing
        end
[...]

PR #271, opened 2018-07-25
WIP Remove Int timestep variant

lrennels:
See issue #247


src/core/types.jl
```diff
@@ -327,7 +327,7 @@ mutable struct ComponentInstance{TV <: ComponentInstanceVariables, TP <: Compone
 
     run_timestep::Union{Void, Function}
     use_integer_time::Bool                  # will be deprecated when switching to timesteps in all cases
```
rjplevin: Comment out `use_integer_time::Bool`, too.

lrennels:
If we want `t` to be automatically assumed to be an `AbstractTimestep` within `run_timestep`, then we need a few changes, some to the API so I want to check in with @davidanthoff and @rjplevin sine these are substantial changes and we may want to think of ways to simplify the API as this introduces opportunities for bugs especially with needing `gettime(ts)` and `ts.t`.  You can see the changes below in the comments.

```
@defcomp Bar2 begin
    inputB = Parameter(index=[time])
    output = Variable(index=[time])
    
    function run_timestep(p, v, d, ts) #ts will be assumed to be an AbstractTimestep
        if gettime(ts) < 2005 #we use gettime(ts) here and interpret it as a year not an index
            v.output[ts] = 0
        else
            v.output[ts] = p.inputB[ts.t-1] * ts.t #we have to use ts.t here
        end
    end
end
```

rjplevin:
As I recall, we discussed adding methods for arithmetic operators to Timestep, which would solve the `ts - 1` problem. I'm less sure what to do about `x * ts`. Is that used in real models, or only in testing?

Our `getindex` methods should handle the case of `p.inputB[ts]`. @davidanthoff - is this consistent with how you were expecting this to work?

lrennels:
Yep, adding arithmetic operators would fix the `ts - 1`.  I started to add those but then went down a bit of a rabbit hole with the cases such as do we want users to have to do `gettime(ts) == 2005` and the `x * t` case so I paused.

lrennels:
Got it, yes I too wonder if that's a real use case, along with something like *x + ts* or *x - ts* ... which would seem silly to me.  It seems that if you wanted to do those you also could just create your own counter variable to use in their place ... but I guess it depends on if these use cases are common at all.  I've never seen them.

rjplevin:
I think this is ready to go once you address merge conflicts. I think once we get all the changes related to Int timesteps in we'll stop having CI errors.

PR #279, opened 2018-07-30
Function Naming

lrennels:
- remove unnecessary terms from `export` list
- apply changes to the naming, following conventions for `!`, `_`, and naming of `component`, `parameter`, and `variable`
- follow through on these changes throughout `Mimi` code base, documentation (including `Porting` doc), and other models


src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
     Model,
-    addcomponent,
-    add_connector_comps,
-    add_dimension!,
-    addparameter,
-    compdef,
-    compdefs,
-    compinstance,
-    compkeys,
+    addcomponent,  # push!, add_component!, add_component, addcomp!, add!; could also have multiple functions like insert!, push! etc. instead of before after keywords
```
lrennels: suggestion:  `add_comp`
reason:  matches `_` convention; matches using `comp, var, and param`; `add!` seems perhaps too simplified for now since we generally do specific comp, var, or param

src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
+    add_connector_comps, #should this be deleted?  Docs indicate we need it ...
```
lrennels: This gets automatically called within each `@defcomp`, but I think we still need it at the end of the API for building models.

src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
     components,
-    connect_parameter,
+    connect_parameter, # connect!, bind, assign, link; ! yes or no; frontrunner: connect!
```
lrennels: suggestion:  `connect_param`
reason:  matches the convention for `!` because this doesn't change something that exists (unless you mean the model itself but that's not generally what we mean with `!`); matches convention of `<action>_<comp/param/var>` that we generally seem to follow

src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
     create_marginal_model,
-    datatype,
-    description,
-    dimension,
-    dimensions,
-    disconnect!,
+    disconnect!, # disconnect_parameter!; frontrunner: disconnect!
```
lrennels: suggestion:  `disconnect_param!`
reason:  matches the above `connect_param` but uses `!` since it alters an existing connection

src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
-    # plot_comp_graph,
-    replace_component,
-    run_timestep,
-    set_dimension!,
-    set_leftover_params!,
-    set_parameter!,
-    setproperty!,
-    unit,
-    variables
+    replace_component, # Add !, comp vs, component, replace!
```
lrennels: suggestion:  `replace_comp!`
reason:  matches `!` convention;  matches aforementioned `<action>_<comp/param/var>` pattern

src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
+    set_dimension!, # Think hard about axis-dimension-index-blabla
```
lrennels: leave this alone for now since this is a bigger design discussion?

src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
+    set_leftover_params!, #rethink in general
```
lrennels: leave this alone for now, we may want to discuss the use cases of this function but I think that's a bigger discussion for later

src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
+    set_parameter!, # Just make parameter match other parts
```
lrennels: suggestion:  `set_param!`
reason:  matches other parts

src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
-    dimension,
-    dimensions,
-    disconnect!,
+    disconnect!, # disconnect_parameter!; frontrunner: disconnect!
     explore,
     getdataframe,
-    getproperty,
     gettime,
     get_parameter_value,
     get_variable_value,
```
lrennels: `get_parameter_value ` -->  `get_param_value` to match `parameter` --> `param`?
`get_variable_value ` -->  `get_var_value` to match `variable` --> `var`?

src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
     hasvalue,
-    interpolate,
     is_first,
     is_last,
     load_comps,
     modeldef,
     name,
     new_component,
```
lrennels: `new_component ` -->  `new_comp` to match `component` --> `comp`?

src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
     parameters,
-    # plot,
-    # plot_comp_graph,
-    replace_component,
-    run_timestep,
-    set_dimension!,
-    set_leftover_params!,
-    set_parameter!,
```
lrennels: Change to `set_param!` event though this will be removed from the API?

src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
     get_variable_value,
     hasvalue,
-    interpolate,
     is_first,
     is_last,
     load_comps,
     modeldef,
     name,
     new_component,
     parameters,
```
lrennels: `parameters ` -->  `params` to match `parameter` --> `param`?

src/Mimi.jl
```diff
@@ -14,50 +14,32 @@ using StringBuilders
...
-    # plot,
-    # plot_comp_graph,
-    replace_component,
```
lrennels: `replace_component ` -->  `replace_comp` to match `component` --> `comp`?

src/Mimi.jl
rjplevin: I agree.

src/Mimi.jl
rjplevin: We could make `add_connector_comps` the first step of `build(m::Model)`. This would allow us to remove it from `@defcomp` and alleviate the user from calling it interactively. Any objections?

src/Mimi.jl
rjplevin: Actually, I'd argue for `!`. I interpret that as indicating a function with side-effects, i.e., not a pure function that returns a value but leaves all else unchanged.

src/Mimi.jl
rjplevin: Agreed.

src/Mimi.jl
rjplevin: Agreed.

src/Mimi.jl
rjplevin: Agreed.

src/Mimi.jl
rjplevin: Agreed.

src/Mimi.jl
lrennels: Ok that might be a more consistent interpretation, maybe above should be `add_comp!` then?

src/Mimi.jl
lrennels: I like that!  I'm not fully comfortable with the internals of editing a built model yet ... but I assume any time a big change is made which requires a new call to `add_connector_comps`, we probably `decache` and then `build` again.  That question doesn't affect this decision, just trying to be clear when I'm editing docs.

src/Mimi.jl
rjplevin: Yes.

lrennels:
@davidanthoff @rjplevin This PR is ready, I have done all the naming changes here, updated the documentation, and updated integration branches.  There are four "TODO" comments in the export list, which I will handle in my next PR since they are slightly more involved and I'd rather keep them separate.

PR #291, opened 2018-08-02
Fixed bug in processing declared type of model Parameters.

rjplevin:
If a default was given as an Int, it would override the declaration of Parameter type, confounding a subsequent MCS which tries to store a float into this location.

### DIFF
--- a/src/core/build.jl
+++ b/src/core/build.jl
@@ -1,7 +1,7 @@
 connector_comp_name(i::Int) = Symbol("ConnectorComp$i")
 
 # Return the datatype to use for instance variables/parameters
-function _instance_datatype(md::ModelDef, def::DatumDef, first::Int)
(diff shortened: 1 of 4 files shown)


rjplevin:
We should add some tests for this...

lrennels:
You're right, we don't want to have to wait for testing dependencies to find a simple bug like that.

rjplevin:
I've updated the treatment of parameter types. FUND now runs completely (hopefully, correctly), as do the tutorial models and the Mimi tests. The way it works now:

- If explicit type is given, that is used.
  - Example: `foo::Vector{Int} = Parameter(index=[:regions])`
- If a default value is given, and no explicit type is given, the type of the default is used.
  - Example: `foo = Parameter(index=[:regions], default=[.234, .234, .3465, .432])`
- The model's default type is assumed (generally Float64)

I haven't tested this with FUND's MCS logic...

PR #295, opened 2018-08-03
Revised treatment of default parameter values and parameter types.

rjplevin:
- Default values must have the same number of dimensions as the parameter (via the indexes=[...] option or lack thereof)
- Default values and other values are converted to the parameter's declared type (or lacking a declaration, to the Model's number_type) when set.
- Added tests for correct treatment of "number_type" applied to params of unspecified types


lrennels:
@ckingdon95 does this look good to you?  If so I think we can merge it.

ckingdon95:
I don't feel fully qualified to review all of the internal changes, I'm just not as familiar with the whole defcomp macro, but the test files looks good to me!

PR #302, opened 2018-08-23
Update comp_defs when redefining time dimension

ckingdon95:
This PR defines a new internal function, reset_run_periods! to be called by set_dimension! if the user is redefining the time index.

reset_run_periods redefines the 'first' and 'last' fields of each component definition to be compatible with the model's new time index, using the already defined function set_run_period! that is originally  called during calls to add_comp!

If the model's new time index is wider than a component definition's current first and last values, this is fine and nothing is changed and the component will run on a shorter time index than the model.

But if the model's new time index is narrower than a component definition's, then this function gives a warning and the first and/or last values of the component are modified.

This also updates the check_labels function to check the length of the component's run period, not the model's time index, when setting a parameter value along the time dimension.

I'm not certain if this is the functionality we want, but I think it is. I ran into a problem when redefining a model's time index to be shorter, then trying to assign parameter values, which gave an (appropriate) error because the component definitions first and last values were still set to the original time dimension's end points. There may be an alternative solution, however, where instead of adding this function, we could not call set_run_period! during the add_comp stage, and instead do set_run_period! during build. I haven't fully explored this option though, and it might not work. (one complication I can think of already is when the user explicitly defines a first and last value when calling add_comp)


src/core/defs.jl
```diff
@@ -187,6 +187,24 @@ dim_value_dict(md::ModelDef) = Dict([name => collect(values(dim)) for (name, dim
 Base.haskey(md::ModelDef, name::Symbol) = haskey(md.dimensions, name)
 
 isuniform(md::ModelDef) = md.is_uniform
+
+
+# Helper function invoked when the user resets the time dimension with set_dimension!
+# This function calls set_run_period on each component definition to reset the first and last values.
```
lrennels: tiny fix to the description ... correct function name above to `set_run_period!`

ckingdon95: will do!

ckingdon95:
@lrennels I'll add tests! Also, does the placement of this function in the defs.jl file make sense? I just put it right above the set_dimension! function that calls it, but I'm not sure if there's a different organization of this file that I should know about

lrennels:
Good question.  Putting it in this file definitely makes sense, so right now you put it in the `#Dimensions` section of the file and called by a function there,  and `set_run_period!` is in `Other` but called within a function in `#Model`.  They certainly both make sense, but I may argue to put it down with `set_run_period!` so at least those two are together?

src/core/connections.jl
```diff
@@ -44,7 +44,16 @@ function _check_labels(md::ModelDef, comp_def::ComponentDef, param_name::Symbol,
...
+            if dim == :time 
+                first = comp_def.first 
+                last = comp_def.last 
+                t = md.dimensions[:time]
+                n1 = t[last] - t[first] + 1
+                n2 = size(ext_param.values)[i]
+                if n1 != n2
+                    error("Mismatched data size for a parameter connection: $(comp_def.comp_id): dim :time has $n1 elements; parameter :$param_name has $n2 elements.")
+                end
+            elseif dim_count(md, dim) != size(ext_param.values)[i]
```
rjplevin: You can consolidate the redundant portions of this so that you capture n1, n2, and dim in both branches of the conditional, then after the conditional, test n1 != n2 and raise the error.

src/core/defs.jl
```diff
@@ -187,6 +187,24 @@ dim_value_dict(md::ModelDef) = Dict([name => collect(values(dim)) for (name, dim
 Base.haskey(md::ModelDef, name::Symbol) = haskey(md.dimensions, name)
 
 isuniform(md::ModelDef) = md.is_uniform
+
+
+# Helper function invoked when the user resets the time dimension with set_dimension!
+# This function calls set_run_period! on each component definition to reset the first and last values.
+function reset_run_periods!(md, first, last)
+
+    for comp_def in values(md.comp_defs)
```
rjplevin: You can use the fn defined in defs.jl:
`compdefs(md::ModelDef) = values(md.comp_defs)`

src/core/defs.jl
```diff
@@ -187,6 +187,24 @@ dim_value_dict(md::ModelDef) = Dict([name => collect(values(dim)) for (name, dim
...
+        if comp_def.first < first
```
rjplevin: I recommend using a functional style when possible to avoid spreading implementation details of types throughout the code. So instead of directly accessing `comp_def.first`, I'd use these functions (already defined):
```
first_period(comp_def::ComponentDef) = comp_def.first
last_period(comp_def::ComponentDef) = comp_def.last
```

The complier will inline these, so it's the same at run-time, but hiding struct details behind a functional interface allows more flexibility if we change the structure at some point.

src/core/defs.jl
```diff
@@ -187,6 +187,24 @@ dim_value_dict(md::ModelDef) = Dict([name => collect(values(dim)) for (name, dim
...
+function reset_run_periods!(md, first, last)
+
+    for comp_def in values(md.comp_defs)
+        if comp_def.first < first 
+            warn("Resetting $(comp_def.name) component's first timestep to $first")
+        end 
+        if comp_def.last > last 
+            warn("Resetting $(comp_def.name) component's last timestep to $last")
+        end
+        set_run_period!(comp_def, max(comp_def.first, first), min(comp_def.last, last))
```
rjplevin: This is admittedly minor, but calling set_run_period! here happens even when no change is necessary (even though we checked this immediately above) and we call `max` and `min` to recover information we also had already. And it's in a loop over all components. You could economize several ways. Here's one:
```
function reset_run_periods!(md, first, last)
    for comp_def in values(md.comp_defs)
        changed = false
        if comp_def.first < first 
            warn("Resetting $(comp_def.name) component's first timestep to $first")
            changed = true
        else
            first = comp_def.first
        end 

        if comp_def.last > last 
            warn("Resetting $(comp_def.name) component's last timestep to $last")
            changed = true
        else
            last = comp_def.last
        end

        if changed
            set_run_period!(comp_def, first, last)
        end
    end

    nothing
end
```

src/core/defs.jl
```diff
@@ -197,6 +215,9 @@ an integer; or to the values in the vector or range if `keys` is either of those
 function set_dimension!(md::ModelDef, name::Symbol, keys::Union{Int, Vector, Tuple, Range})    
     if haskey(md, name)
```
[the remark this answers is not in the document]
ckingdon95: then won't I have to test haskey(md, name) twice? like this:

```
if name == :time 
        if haskey(md, name)
            warn("Redefining dimension :$name")
            reset_run_periods!(md, keys[1], keys[end])
        end
        md.is_uniform = isuniform(keys)
else
        if haskey(md, name)
            warn("Redefining dimension :$name")
        end
end
```
or am I totally missing a different way to do that? I guess I could do:
```
if dim == :time
    reset_run_periods!(md, keys[1], keys[end])
    md.is_uniform = isuniform(keys)
end
if haskey(md, name)
    warn("Redefining dimension :$name")
end
```

src/core/defs.jl
```diff
@@ -195,12 +220,13 @@ Set the values of `md` dimension `name` to integers 1 through `count`, if `keys`
 an integer; or to the values in the vector or range if `keys` is either of those types.
 """
 function set_dimension!(md::ModelDef, name::Symbol, keys::Union{Int, Vector, Tuple, Range})    
+    if name == :time
+        reset_run_periods!(md, keys[1], keys[end])
+        md.is_uniform = isuniform(keys)
+    end
     if haskey(md, name)
         warn("Redefining dimension :$name")
     end
```
ckingdon95: @rjplevin I tried to simplify this one like you recommended by having name==:time just once, however unless I want to have another separate call to haskey() within this conditional, it will call reset_run_periods! even if the model doesn't already have a time index. I think this is okay, though, because if it doesn't already have a time index, then it doesn't have any components, so reset_run_periods! will loop over an empty dictionary. It still seems a bit inelegant though..

rjplevin: How about:
```
function set_dimension!(md::ModelDef, name::Symbol, keys::Union{Int, Vector, Tuple, Range})
    redefined = haskey(md, name)
    if redefined
        warn("Redefining dimension :$name")
    end

    if name == :time
        md.is_uniform = isuniform(keys)
        if redefined:
            reset_run_periods!(md, keys[1], keys[end])
    end
    
    dim = Dimension(keys)
    md.dimensions[name] = dim
    return dim
end
```

ckingdon95: ah so clever, thank you!

rjplevin: Whoops... some Python syntax snuck in my code.

lrennels:
@ckingdon95 We can probably merge this PR now right?  That way you don't have to deal with it's changes propagating through into the changes shown in your new two branches?

ckingdon95:
@lrennels @rjplevin do you think it's okay for me to merge this without David's review? Also that these changes might become irrelevant after we potentially redesign how components' first and last values get set? (I think it's okay to merge now, cause technically this is a fix for a current bug, but want to check)

lrennels:
I think it's ready to go, yep.

PR #305, opened 2018-08-29
update_param!

ckingdon95:
new functionality:

if `update_timesteps == true`:
- ERROR if it's a scalar parameter (cannot update timesteps for scalars)
- ERROR if the existing param is an Array but not a timestep array
- ERROR if the new values are the wrong size, or a different data type that can't be converted

In the case of multiple updates from a dictionary using `update_params!` if `update_timesteps==true`, then it will not error in the scalar or non-timestep array cases, allowing the user to have a dictionary of all types of parameters.

questions:
- [x] in the case of update_timesteps == true, should we instead be checking the size of the new data against the model's new time index, rather than against the old value's size?
- [x] do we like how explicit the name `update_external_param!` is, or should we make it more consistent and call it `update_param!` instead?
- [x] - I put the calls to type Model in the src/core/model.jl file, and the calls to type ModelDef and the internal supporting functions in src/core/connections.jl. Does this organization make sense?
- [x] - for now i've exported both `update_external_param!` (for a single parameter) and `update_external_params!` (for multiple parameter update from a dictionary). is this okay?
- [x] - I added a new function `get_timestep_array` very similar to the already existing function `get_timestep_instance` that generates a generic-size array, instead of a vector or matrix. Should this function be used instead everywhere? or do we want to keep using vectors and matrices for 1 and 2 dimensional cases?

### DIFF
--- a/docs/src/integrationguide.md
+++ b/docs/src/integrationguide.md
@@ -105,7 +105,7 @@ The full API:
 
 ### Updating an external parameter
 
-The function `update_external_parameter` is now written as `update_external_param`.
+The function `update_external_parameter` is now written as `update_external_param!`, and `update_external_parameters` is now written as `udpate_external_params!`.  There is also a new optional argument `update_timesteps` which can be used if the external parameter being set has a `:time` dimension, and not only values pertaining to those times, but also time keys that are connected to those values have changed.
 
 ### Setting parameters with a dictionary
 
--- a/docs/src/integrationguide.md
+++ b/docs/src/integrationguide.md
@@ -105,7 +105,7 @@ The full API:
 
 ### Updating an external parameter
 
-The function `update_external_parameter` is now written as `update_external_param!`, and `update_external_parameters` is now written as `udpate_external_params!`.  There is also a new optional argument `update_timesteps` which can be used if the external parameter being set has a `:time` dimension, and not only values pertaining to those times, but also time keys that are connected to those values have changed.
+The function `update_external_parameter` is now written as `update_param!`, and `update_external_parameters` is now written as `udpate_params!`.  There is also a new optional argument `update_timesteps` which can be used if the external parameter being set has a `:time` dimension, and not only values pertaining to those times, but also time keys that are connected to those values have changed.
 
 ### Setting parameters with a dictionary
 
--- a/docs/src/integrationguide.md
+++ b/docs/src/integrationguide.md
@@ -105,11 +105,17 @@ The full API:
 
 ### Updating an external parameter
 
-The function `update_external_parameter` is now written as `update_param!`, and `update_external_parameters` is now written as `udpate_params!`.  There is also a new optional argument `update_timesteps` which can be used if the external parameter being set has a `:time` dimension, and not only values pertaining to those times, but also time keys that are connected to those values have changed.
+To update an external parameter, use the functions `update_param!` and `udpate_params!` (previously known as `update_external_parameter` and `update_external_parameters`, respectively.)  Their calling signatures are:
+
+*  `update_params!(md::ModelDef, parameters::Dict; update_timesteps = false)`
+
+* `update_param!(md::ModelDef, name::Symbol, value; update_timesteps = false)`
+
+For external parameters with a `:time` dimension, passing `update_timesteps=true` indicates that the time _keys_ (i.e., year labels) should also be updated in addition to updating the parameter values.
 
 ### Setting parameters with a dictionary
 
-The function `setleftoverparameters` is now written as `set_leftover_params!`.
+The function `set_leftover_params!` replaces the function `setleftoverparameters`.
 
 ### Using NamedArrays for setting parameters
 
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -30,5 +30,6 @@ set_leftover_params!
 setproperty!
 set_param! 
 variables  
+update_external_param!
 update_external_params!
 ```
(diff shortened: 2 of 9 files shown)


ckingdon95:
note to self, still need to: add tests for multiple updates function

davidanthoff:
I'd almost be tempted to make these warnings errors. After all, one is calling a function with something that doesn't make sense, right?

A more generic question is that I don't fully understand the use case here :) Do I understand this correctly, that a call to update an external parameter in a model here would actually change the array that we pass to the model? Why?

lrennels:
Some responses to Cora's questions, @rjplevin @davidanthoff feel free to add.

- in the case of update_timesteps == true, should we instead be checking the size of the new data against the model's new time index, rather than against the old value's size?

So in this case, there is both new data and new timestep labels, correct?  If so, I think yes the check should be against the new data otherwise we'll hit unwanted errors if the size of the time index has changed.

- do we like how explicit the name `update_external_param!` is, or should we make it more consistent and call it `update_param!` instead?

Hm that's a good question, we don't have a matching `update_internal_param!` functionality, which would be one reason to keep `update_external_param!`.  I'd probably vote to keep `external` in the name, to prevent users (and us) from thinking we can apply this to internal parameters, but don't have a strong opinion.

- I put the calls to type Model in the src/core/model.jl file, and the calls to type ModelDef and the internal supporting functions in src/core/connections.jl. Does this organization make sense?

I need to think on this one, but at first glance that seems to make sense.

- for now i've exported both `update_external_param!` (for a single parameter) and `update_external_params!` (for multiple parameter update from a dictionary). is this okay?

I think that's fine, if possible we could try to create one function with two methods, but they are so different in terms of arguments and functionality that it might be best to keep them separate.

- I added a new function `get_timestep_array` very similar to the already existing function `get_timestep_instance` that generates a generic-size array, instead of a vector or matrix. Should this function be used instead everywhere? or do we want to keep using vectors and matrices for 1 and 2 dimensional cases?

I think we had performance in mind when using vectors and matrices for several of these functions, but I also need to look at this one more closely.

lrennels:
I'd agree with @davidanthoff that those warnings might belong as errors.

ckingdon95:
@davidanthoff @lrennels 
The warnings instead of errors were so that when performing multiple updates from a dictionary, using `update_external_params!`, the user could use the `update_timestep` argument for parameters that need it, and that scalar parameters or non-timestep array parameters could still be in the same dictionary without causing an error. I agree though that it seems a bit odd to allow this behavior. 

If alternatively we turn them into errors, this prevents a user from having one update dictionary, and they would have to split it into two calls, one with `update_steps=true` and one with `=false`. 

To your more generic question David, originally this function `update_external_param` was used to change the data values of an external parameter, without making a new array. But now if the user also wants to update the timesteps with `update_timesteps=true`, then we need to build a new TimestepArray with the correct time labels. is that what you were asking? (For context, the use case here is with the IWG version of PAGE, where for each scenario I need to replace external data values, and also update the time labels of the external parameters)

ckingdon95:
@lrennels regarding the name of the function: I'm leaning more towards `update_param!`, since users never really use the term "external paramater" anywhere, and I can't think of there ever being an `update_internal_param!` since users can already do that with calls to `connect_param`.

rjplevin:
FWIW, I agree with @cora on the naming.

lrennels:
I have been overruled!  Fine by me to change it, we'll just want to make sure it gets changed in all docs etc. as well as integration branches.

lrennels:
@ckingdon95 Ah now I understand that design decision better, so this would be an issue for `update_external_params!` where you are doing multiple calls to `update_external_param!`.  In that case I'd lean towards leaving it the way you've done it, but I can think about it.

rjplevin:
@ckingdon95 wrote
>> I added a new function get_timestep_array very similar to the already existing function get_timestep_instance that generates a generic-size array, instead of a vector or matrix. Should this function be used instead everywhere? or do we want to keep using vectors and matrices for 1 and 2 dimensional cases?

@lrennels wrote:
> I was not part of the original decision to split up the vectors and matrices, though we've discussed whether that is necessary or causes confusing redundancy. Originally I think we had performance in mind when we made specific methods for vectors and arrays, though that's probably not relevant for the get... methods. What's the need for get_timestep_array? Were there cases of over 2 dimensions?

The reason for this is that the most common cases have 1 or 2 dims, and the _n_-dimensional version had some inefficiencies (e.g., building a list of dimensions) that can be avoided in the simpler cases. We have not benchmarked this though, so it may be a "nothing burger" as they say. Might be worth constructing a benchmarking case to see if we're really benefiting here. Simplifying it would be nice!

lrennels:
@rjplevin I love the "nothing burger" term ... @ckingdon95 perhaps for now leave it as is and then we can add another issue to benchmark the differences?  That's probably not necessary if we're just talking about this `get_timestep...` function in isolation, but an overhaul to remove the `vector` and `matrix` functions would be it's own PR.

rjplevin:
Good idea. I'll create the issue now.

ckingdon95:
@rjplevin @lrennels 
Agreed that doing an overall examination of Timestep vectors/matrices/arrays should be a separate PR. 
So for doing my updates here in the case where I need to build a new array, should I default to using vectors and matrices for 1 and 2 dimensional cases, and have the fallback to the generic function I wrote? Or do you think I should delete my generic function for now, just use the existing `get_timestep_instance` function, which will error for higher dimension cases?

lrennels:
I think the former seems fine, unless there was a specific reason that we wanted an error on the higher dimension cases ... but I can't think of one!

Issue #289, opened 2018-08-02
Make replace_comp! re-connect parameters


ckingdon95:
Do we want the replace_comp! function to re-connect any previously existing parameter connections to and from the component that's replaced?

rjplevin:
I'd say yes, though I'd make it an optional argument to allow this behavior to be suppressed.

ckingdon95:
other note for when I get around to working on replace_comp: I think the new component should get added in the same position in the component list as the old one, without requiring that the user specify a before or after argument

rjplevin:
> Do we want the replace_comp! function to re-connect

How about an optional argument `reconnect::Bool=true`?

ckingdon95:
PR #307

PR #307, opened 2018-08-30
WIP replace_comp!

ckingdon95:
Address issue #289 , with optional bool argument `reconnect`.

If reconnect==false, same behavior as before.

If reconnect==true, then replace_comp! only deletes the old component from md.comp_defs, and does not delete the internal or external parameter connections that the component belongs to.

It performs the following checks (in the case of reconnect==true):
- it checks that the new component definition has any parameter names that are present in any incoming internal parameter connections, and checks that the dimensions and datatype are the same. Question: should all of the datum fields need to match? Like units and description?
- it checks that the new component definition has any variable names that are present in any outgoing internal parameter connections, and checks that the dimensions and datatype are the same.
- it checks if the new component has the parameter names that are present in any external parameter connections. If it's missing a parameter, it just gives a warning and removes the external parameter connection (should this error?). But if it has the parameter and has different dimensions or datatypes, then it errors

### DIFF
--- a/src/core/defs.jl
+++ b/src/core/defs.jl
@@ -539,11 +539,16 @@ end
 """
     replace_comp!(md::ModelDef, comp_id::ComponentId, comp_name::Symbol=comp_id.comp_name;
         first::VoidInt=nothing, last::VoidInt=nothing,
-        before::VoidSymbol=nothing, after::VoidSymbol=nothing)
+        before::VoidSymbol=nothing, after::VoidSymbol=nothing,
+        reconnect::Bool=true)
 
-Replace the component with name `comp_name` in model `md`  with the component
-`comp_id` using the same name.  The component is added at the end of 
-the list unless one of the keywords, `first`, `last`, `before`, `after`.
+Replace the component with name `comp_name` in model definition `md` with the 
+component `comp_id` using the same name. The component is added in the same 
+position as the old component, unless one of the keywords `before` or `after` 
+is specified. The component is added with the same first and last values, 
+unless the keywords `first` or `last` are specified. Optional boolean argument 
+`reconnect` with default value `true` indicates whether the existing parameter 
+connections should be maintained in the new component.
 """
 function replace_comp!(md::ModelDef, comp_id::ComponentId, comp_name::Symbol=comp_id.comp_name;
                            first::VoidInt=nothing, last::VoidInt=nothing,
--- a/src/core/defs.jl
+++ b/src/core/defs.jl
@@ -554,24 +554,31 @@ function replace_comp!(md::ModelDef, comp_id::ComponentId, comp_name::Symbol=com
         error("Cannot replace '$comp_name'; component not found in model.")
     end
 
-    # TODO: Get original position if new before or after not specified
-    # if before == nothing && after == nothing
-    #     comps = collect(keys(md.comp_defs))
-    #     next_idx = findfirst(comps, comp_name) + 1
-    #     before = comps[next_idx]
-    # end 
+    # Get original position if new before or after not specified
+    if before == nothing && after == nothing
+        comps = collect(keys(md.comp_defs))
+        n = length(comps)
+        if n > 1
+            idx = findfirst(comps, comp_name)
+            if idx == n 
+                after = comps[idx - 1]
+            else
+                before = comps[idx + 1]
+            end
+        end
+    end 
 
     # Get original first and last if new run period not specified
-    first = first == nothing ? md.comp_defs[comp_name].first : first
-    last = last == nothing ? md.comp_defs[comp_name].last : last
+    old_comp = md.comp_defs[comp_name]
+    first = first == nothing ? old_comp.first : first
+    last = last == nothing ? old_comp.last : last
 
     if reconnect
         # Assert that new component definition has same parameters and variables needed for the connections
 
-        old_comp = md.comp_defs[comp_name]
         new_comp = compdef(comp_id)
 
-        function compare_datum(dict1, dict2)
+        function _compare_datum(dict1, dict2)
             set1 = Set([(k, v.datatype, v.dimensions) for (k, v) in dict1])
             set2 = Set([(k, v.datatype, v.dimensions) for (k, v) in dict2])
             return set1 >= set2
@@ -582,7 +589,7 @@ function replace_comp!(md::ModelDef, comp_id::ComponentId, comp_name::Symbol=com
         param_filter = (k, v) -> k in incoming_params
         old_params = filter(param_filter, old_comp.parameters)
         new_params = new_comp.parameters
-        if !compare_datum(new_params, old_params)
+        if !_compare_datum(new_params, old_params)
             error("Cannot replace and reconnect; new component does not contain the same definitions of necessary parameters.")
         end
         
@@ -591,7 +598,7 @@ function replace_comp!(md::ModelDef, comp_id::ComponentId, comp_name::Symbol=com
         var_filter = (k, v) -> k in outgoing_vars
         old_vars = filter(var_filter, old_comp.variables)
         new_vars = new_comp.variables
-        if !compare_datum(new_vars, old_vars)
+        if !_compare_datum(new_vars, old_vars)
             error("Cannot replace and reconnect; new component does not contain the same definitions of necessary variables.")
         end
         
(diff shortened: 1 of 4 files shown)


ckingdon95:
still need to:
-get before/after
-get first/lasts
-clean up parameter connection checking code

src/core/defs.jl
```diff
@@ -547,9 +547,82 @@ the list unless one of the keywords, `first`, `last`, `before`, `after`.
...
+        datum_map = kv -> (kv[1], kv[2].datatype, kv[2].dimensions)  # Want to compare datum names, type, and dimensions as a tuple. Should we compare all values of the DatumDef?
+
+        # Check incoming parameters
+        incoming_params = map(ipc -> ipc.dst_par_name, internal_param_conns(md, comp_name))
+        param_filter = (k, v) -> k in incoming_params
+        old_params = filter(param_filter, old_comp.parameters)
+        new_params = new_comp.parameters
+        old_set = Set(map(datum_map, collect(old_params)))
+        new_set = Set(map(datum_map, collect(new_params)))
+        if !(new_set >= old_set)
```
rjplevin: You can eliminate some of the redundant code here and below that maps `datum_map` over two vectors, converts to sets, and checks the subset. Also, no need to `collect` a vector in a `map`.

src/core/defs.jl
```diff
@@ -547,9 +547,82 @@ the list unless one of the keywords, `first`, `last`, `before`, `after`.
...
+    # last = last == nothing? md.comp_defs[comp_name].last : last
+
+    if reconnect
+        # Assert that new component definition has same parameters and variables needed for the connections
+
+        old_comp = md.comp_defs[comp_name]
+        new_comp = compdef(comp_id)
+
+        # datum_map = (k, v) -> (k, v.datatype, v.dimensions)  # Want to compare datum names, type, and dimensions as a tuple. Should we compare all values of the DatumDef?
+        datum_map = kv -> (kv[1], kv[2].datatype, kv[2].dimensions)  # Want to compare datum names, type, and dimensions as a tuple. Should we compare all values of the DatumDef?
```
rjplevin: Curious why the more complicated `kv` version is used instead of `(k, v)`. Do some values have length != 2?

src/core/defs.jl
```diff
@@ -547,9 +547,82 @@ the list unless one of the keywords, `first`, `last`, `before`, `after`.
...
-                           before::VoidSymbol=nothing, after::VoidSymbol=nothing)
-    delete!(md, comp_name)
+                           before::VoidSymbol=nothing, after::VoidSymbol=nothing,
+                           reconnect::Bool=true)
+
+    if ! haskey(md.comp_defs, comp_name)
+        error("Cannot replace '$comp_name'; component not found in model.")
+    end
+
+    # Get original position if new before or after not specified
```
rjplevin: Is this commented out because it's WIP? Add "TODO:" to the comments to indicate this. In general, we try to avoid checking in commented out code since it's easy to lose track of when the author intends to use it vs was just nervous about deleting it. (We have git for that!)

src/core/defs.jl
```diff
@@ -547,9 +547,82 @@ the list unless one of the keywords, `first`, `last`, `before`, `after`.
...
+        old_vars = collect(filter(var_filter, old_comp.variables))
+        new_vars = collect(new_comp.variables)
+        if !(Set(map(datum_map, new_vars)) >= Set(map(datum_map, old_vars)))
+            error("Cannot replace and reconnect; new component does not contain the same definitions of necessary variables.")
+        end
+        
+        # Check external parameter connections; don't error in this case? just warn?
+        remove = []
+        for epc in external_param_conns(md, comp_name)
+            p = epc.param_name
```
rjplevin: Better to call this `param_name`. Just `p` makes it sound like it's a parameter object (as in `new_p` and `old_p` below.)

test/test_replace_comp.jl
```diff
@@ -0,0 +1,111 @@
+module TestReplaceComp
```
rjplevin: Nice!

src/core/defs.jl
ckingdon95: yes this is still WIP. I put a todo list above in the comments, but I'll start adding it in the code as well!

ckingdon95:
need to: confirm that the behavior in the case of external parameters is what we want

src/core/defs.jl
```diff
@@ -539,17 +539,104 @@ end
...
+            error("Cannot replace and reconnect; new component does not contain the same definitions of necessary variables.")
+        end
+        
+        # Check external parameter connections
+        remove = []
+        for epc in external_param_conns(md, comp_name)
+            param_name = epc.param_name
+            if ! haskey(new_params, param_name)  # TODO: is this the behavior we want? don't error in this case? just warn?
+                warn("Removing external parameter connection from component $comp_name; parameter $param_name no longer exists in component.")
+                push!(remove, epc)
```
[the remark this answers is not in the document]
lrennels: This is a good question.  As you already implemented with errors, we certainly don't want to make any connections that don't match up, but this is a middle-ground case.  I'm not familiar with the use cases here, so in my head you would only want to `replace_comp!` with a very similar component that uses all the same external parameters, and thus throwing an error could make sense.  That said, I could see using the function to replace a function with a **very similar** one that different on one or two parameters and thus the warning would be sufficient.  So I guess it depends on use cases?

src/core/defs.jl
```diff
@@ -539,17 +539,104 @@ end
...
+            return set1 >= set2
+        end
+
+        # Check incoming parameters
+        incoming_params = map(ipc -> ipc.dst_par_name, internal_param_conns(md, comp_name))
+        param_filter = (k, v) -> k in incoming_params
+        old_params = filter(param_filter, old_comp.parameters)
+        new_params = new_comp.parameters
+        if !_compare_datum(new_params, old_params)
+            error("Cannot replace and reconnect; new component does not contain the same definitions of necessary parameters.")
```
lrennels: I was wondering if we wanted to be more specific about which doesn't match (I'm trying to keep an eye on our Error messages in general to improve them), but I think doing so would open up more cases and logic than is worth it here.  So no need to make a change, just wanted to mention that.

src/core/defs.jl
```diff
@@ -539,17 +539,104 @@ end
...
+        end
+        
+        # Check outgoing variables
+        outgoing_vars = map(ipc -> ipc.src_var_name, filter(ipc -> ipc.src_comp_name == comp_name, md.internal_param_conns))
+        var_filter = (k, v) -> k in outgoing_vars
+        old_vars = filter(var_filter, old_comp.variables)
+        new_vars = new_comp.variables
+        if !_compare_datum(new_vars, old_vars)
+            error("Cannot replace and reconnect; new component does not contain the same definitions of necessary variables.")
```
lrennels: Same comment as above.

src/core/defs.jl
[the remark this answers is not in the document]
lrennels: I agree that this is worth discussion, I added a comment in the general `PR` thread.

test/test_replace_comp.jl
```diff
@@ -0,0 +1,132 @@
...
+
```
lrennels: Nice job on these they look organized and comprehensive!

src/core/defs.jl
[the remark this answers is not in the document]
ckingdon95: For more context on use cases: so far when I've had to replace components for the IWG's versions of the models, there have been 4 different cases:
1. The new component has all the same parameters and variables, there is just a change in the run_timestep function (so we want to reconnect everything, and there should be no errors)
2. The new component has additional parameters and/or variables (still everything else should get reconnected and no errors)
3. The new component now endogenously calculates something that was previously an external parameter (parameter `:p` is now a variable `:p`), but all other internal parameter connections are the same. This is the case where I want to use `reconnect=true`, but to just warn on the disappearance of an external parameter
4. The new component now has exogenous values for a parameter that was previously an internal connection from a different component. This is the case where we would use `reconnect=false` and set  the parameter to the new external value.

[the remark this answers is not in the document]
lrennels: I think we should leave it this way, with a warning but no error.  It seems to important to err on the side of caution when we are making connections based on assumptions, but in this case we are just warning of a disappearance which makes sense to me.  @rjplevin what do you think?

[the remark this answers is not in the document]
rjplevin: Agreed!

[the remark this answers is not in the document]
davidanthoff: Yes, agreed. Lets try it with this, and if we don't like it, we can still change it later.

ckingdon95:
@rjplevin I think I've implemented all of your requested changes, but is there anything else I should change before merging this one?

rjplevin:
Should be good to merge.

PR #311, opened 2018-09-04
Fix generate trials bug

ckingdon95:


ckingdon95:
@rjplevin does that make sense to you that it was just the `Base.done(iter, idx)` line that needed changing and not the `Base.done(mcs::MonteCarloSimulation, trialnum)`? I don't fully understand the iterator methods being used here so I'm not sure why this worked

Issue #312, opened 2018-09-05
mcs EmpiricalDistribution bug


ckingdon95:
PR #315 

I haven't tried to figure out why this is happening yet, but unlike other random variable distributions, EmpiricalDistributions fail to re-use the same trial values across different scenarios. For example, the following script produces the output below, where the Normal variable has the same values across the different scenarios, and the EmpricialDistribution has new values each time:

```
# Show EmpiricalDistribution bug

using Mimi
using Distributions

csvalues = collect(1:10)
csprobs = 0.1 * ones(10)

mcs_test = @defmcs begin
    cs = EmpiricalDistribution(csvalues, csprobs)
    n = Normal()
end 

@defcomp test begin
    cs = Parameter(default = 5)
    n = Parameter(default = 0)
    function run_timestep(p, v, d, t) end
end

scenario_args = [
    :num => [1,2,3,4]
]

function scenario_func(mcs::MonteCarloSimulation, tup::Tuple)
    println("Scenario $(tup[1])")
end

function post_trial_func(mcs::MonteCarloSimulation, trialnum::Int, ntimesteps::Int, tup::Tuple)
    m, = mcs.models
    println("   trial $trialnum: normal=", m[:test, :n], ", cs=", m[:test, :cs])
end

m = Model()
set_dimension!(m, :time, 2000:2001)
add_comp!(m, test)
set_model!(mcs_test, m)

trials = 2
generate_trials!(mcs_test, trials; filename = "trials.csv", sampling=RANDOM)

run_mcs(mcs_test, trials; 
        output_dir = "out",
        scenario_args = scenario_args,
        scenario_func = scenario_func, 
        post_trial_func = post_trial_func
        )
```

This prints the following results (and saves it as output):

```
Scenario 1
   trial 1: normal=-0.22451710040006626, cs=10.0
   trial 2: normal=-0.049632423579759284, cs=5.0
Scenario 2
   trial 1: normal=-0.22451710040006626, cs=4.0
   trial 2: normal=-0.049632423579759284, cs=2.0
Scenario 3
   trial 1: normal=-0.22451710040006626, cs=7.0
   trial 2: normal=-0.049632423579759284, cs=9.0
Scenario 4
   trial 1: normal=-0.22451710040006626, cs=3.0
   trial 2: normal=-0.049632423579759284, cs=7.0
```

ckingdon95:
I'm going to try to debug this myself, but wanted to track it as a public issue for now

[...]

PR #314, opened 2018-09-05
Make SampleStores for EmpiricalDistributions during generate_trial

ckingdon95:
address issue #312 
either by make the rand! function loop over everything in the rvdict (instead of just the dist_rv list), OR we need to make EmpricialDistributions part of the dist_rv list

TODO: add tests for whichever functionality we choose

### DIFF
--- a/src/mcs/montecarlo.jl
+++ b/src/mcs/montecarlo.jl
@@ -204,9 +204,9 @@ function Base.rand!(mcs::MonteCarloSimulation)
     rvdict = mcs.rvdict
     trials = mcs.trials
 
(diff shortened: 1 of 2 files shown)


ckingdon95:
@rjplevin I think issue #312 is the easier one to fix first, but we also need to address the other considerations you brought up about whether EmpiricalDistributions should be in the dist_rv list.

ckingdon95:
bigger question is: do EmpricalDistributions belong in the dist_rv list, and if so, what is the purpose of this list? can there be any RV's that aren't distributions?

PR #315, opened 2018-09-05
Enable automatic storage of EmpiricalDistribution values

rjplevin:
I removed the type restrictions on `dist_rvs` and this seems to have solved the original problem. Not sure we have enough tests to verify that it doesn't create other problems...

Issue #312


rjplevin:
Sure; more tests are always welcome!

> Just wondering what the original intention of having the dist_rvs list was

Me, too! I don't remember; the test might be an artifact.

ckingdon95:
@rjplevin before merging this, I just wanted to confirm the question I left above, that it still makes sense for a MonteCarloSimulation type to have a separate dist_rv list and an rvdict, and that there wouldn't be any cases that a random variable doesn't come from a distribution defined in the Distribution package or an EmpiricalDistribution.

Issue #306, opened 2018-08-30
Complications with resetting the time dimension 


ckingdon95:
Being addressed with PR #320 

I'm currently working on a modified version of PAGE. I start by getting the original version of page, then start making modifications to it, which include resetting the time dimension, replacing certain components, and updating certain parameter values. 

**Problem with components:**

The components that already exist have a first and last value of 2009 and 2200(the original time index). The new time index is 2010 to 2300. When I update the time index, the first values of the components get updated to 2010, and the last values stay the same at 2200 (because of the new functionality I implemented with the reset_run_periods function). But, this doesn't allow the old components to run on the full new time index, because the CompDef's stored value of 'last' says that the component can't run past 2200.  I think this might warrant a change to how the component first and last values get set. I think it could go like this:

- CompDef's should only store first and last values if the user explicitly gave a first and last value during calls to add-comp!. Otherwise, we should save them as 'nothing'
- Then during the build function, ComponentInstance's would get their first and last values set to whatever the model's time index is at that point, or if the user had explicitly specified first and last values that get stored in the CompDef, then set them to those. 

I think this makes the most sense because Component Definitions are supposed to be a more loose framework, not necessarily tied to a specific start and end value (unless explicitly defined by the user), and they should easily adapt to updates to the Model's dimensions, and then the Component Instance's get tied down with whatever the Model's dimensions are when they get instantiated.

Alternatively, if we really need CompDef's to not have 'nothing' stored in the first and lasts, then I think we need to provide some sort of optional "force_comps" flag in calls to set_dimension! that force components onto the new model time dimension (instead of carefully checking if the new range is wider or narrower than the comp_defs run period like rest_run_periods! does now). 

**Problem with external parameters:**

External parameters that were already defined in the model remain as Variable TimestepArrays with time labels of the old model index. It seems to me that if a user update's the time index, there should be an option for resetting the TimestepArrays that hold external parameter data. This raises questions though of how we want Mimi to interpret changes to the time dimension. If the user had originally provided external data on the old time steps, we probably shouldn't automatically convert it to the new timesteps. But maybe we can provide an option for this. 

For my purposes with page, the external parameters I want to update also involve updating the actual values stored. I have a PR open right now to improve the update_external_param! function to allow updates not just to the values stored, but also to the TIMES of the TimestepArrays that the values are stored in.

rjplevin:
I agree that this suggests some redesign is needed. I'll look at it later today...

[...]

PR #320, opened 2018-09-13
WIP component run periods

ckingdon95:
issue #306 

Changes:
- Component definitions have their first and last values set to `nothing`, unless calls to `add_comp!` explicitly give values
- Component instances get their first and last values set to the first and last values from the component definitions if they are not `nothing`, or from the model's time index at that point
- This involved modifying the ComponentInstance constructor to explicitly take first and last values
- additional internal helper functions were added:
```
first_period(md::ModelDef, comp_def::ComponentDef) = first_period(comp_def) == nothing ? time_labels(md)[1] : first_period(comp_def)
last_period(md::ModelDef, comp_def::ComponentDef) = last_period(comp_def) == nothing ? time_labels(md)[end] : last_period(comp_def)
```
- most internal uses of the functions `first_period(comp_def)` and `last_period(comp_def)` are now replaced with `first_period(comp_def, md)` and `last_period(comp_def, md)` (i.e. in `set_param!` and `_check_labels`)
- new explicit testing of this functionality is found in "test/test_components.jl", but testing in a few other test files also had to be modified

Questions:

- I didn't use the constant `VoidInt` type in the CompDef struct definition because it's defined in a file that's loaded after types.jl get's loaded, and I felt like it was fine to just use the explicit `Union{Void, Int}` here, but let me know if I should move the constant definition into the types.jl file so that it gets used here

### DIFF
--- a/src/core/build.jl
+++ b/src/core/build.jl
@@ -36,7 +36,7 @@ function _vars_type(md::ModelDef, comp_def::ComponentDef)
     var_defs = variables(comp_def)    
     vnames = Tuple([name(vdef) for vdef in var_defs])
     
-    first = first_period(comp_def, md)
+    first = first_period(md, comp_def)
     vtypes = Tuple{[_instance_datatype(md, vdef, first) for vdef in var_defs]...}
 
     return ComponentInstanceVariables{vnames, vtypes}
@@ -72,7 +72,7 @@ Return the resulting ComponentInstance.
 """
 function _instantiate_component_vars(md::ModelDef, comp_def::ComponentDef)
     comp_name = name(comp_def)
-    first = first_period(comp_def, md)
+    first = first_period(md, comp_def)
     vtype = _vars_type(md, comp_def)
     
     vals = [_instantiate_datum(md, def, first) for def in variables(comp_def)]
@@ -161,8 +161,8 @@ function build(md::ModelDef)
         ptypes = Tuple{map(typeof, pvals)...}
         pars = ComponentInstanceParameters{pnames, ptypes}(pvals)
 
-        first = first_period(comp_def, md)
-        last = last_period(comp_def, md)
+        first = first_period(md, comp_def)
+        last = last_period(md, comp_def)
 
         ci = ComponentInstance{typeof(vars), typeof(pars)}(comp_def, vars, pars, first, last, comp_name)
         add_comp!(mi, ci)
--- a/src/core/build.jl
+++ b/src/core/build.jl
@@ -36,7 +36,7 @@ function _vars_type(md::ModelDef, comp_def::ComponentDef)
     var_defs = variables(comp_def)    
     vnames = Tuple([name(vdef) for vdef in var_defs])
     
-    first = comp_def.first
+    first = first_period(comp_def, md)
     vtypes = Tuple{[_instance_datatype(md, vdef, first) for vdef in var_defs]...}
 
     return ComponentInstanceVariables{vnames, vtypes}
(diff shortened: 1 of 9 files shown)


rjplevin:
> should the "unset" first and last values be zero, as they are now, or should I change it to "nothing"? I think "nothing" might be better but haven't fully explored this

In general, I prefer the explicit `nothing`, but in this case, I think 0 is a reasonable and unambiguous sentinel value, and it avoids declaring the type `Union{Void, Int}`.

My preference, though is to avoid the "double negative" (if not X then Y else [if not not X] then Z) and to write:
```
first_period(comp_def) == 0 ? time_labels(md)[1] : first_period(comp_def)
```
instead of 
```
first_period(comp_def) != 0 ? first_period(comp_def) : time_labels(md)[1]
```

src/core/build.jl
```diff
@@ -36,7 +36,7 @@ function _vars_type(md::ModelDef, comp_def::ComponentDef)
     var_defs = variables(comp_def)    
     vnames = Tuple([name(vdef) for vdef in var_defs])
     
-    first = comp_def.first
+    first = first_period(comp_def, md)
```
rjplevin: Pretty much all the other functions that take a `Model`, `ModelInstance` or `ModelDef` have this as the first argument, so I'd make this `first_period(md, comp_def)` for consistency.

src/core/defs.jl
```diff
@@ -194,13 +196,13 @@ isuniform(md::ModelDef) = md.is_uniform
 function reset_run_periods!(md, first, last)
     for comp_def in compdefs(md)
         change = false
-        if first_period(comp_def) < first 
+        if first_period(comp_def) != 0 && first_period(comp_def) < first
```
rjplevin: Can be written 
```
 if 0 < first_period(comp_def) < first 
```

src/core/defs.jl
```diff
@@ -194,13 +196,13 @@ isuniform(md::ModelDef) = md.is_uniform
...
         change = false
-        if first_period(comp_def) < first 
+        if first_period(comp_def) != 0 && first_period(comp_def) < first 
             warn("Resetting $(comp_def.name) component's first timestep to $first")
             change = true
         else
             first = first_period(comp_def)
         end 
-        if last_period(comp_def) > last 
+        if last_period(comp_def) != 0 && last_period(comp_def) > last
```
rjplevin: Assuming `last` is non-negative (it is), this is equivalent to:
```
 if last_period(comp_def) > last
```

src/core/defs.jl
```diff
@@ -485,14 +486,14 @@ function add_comp!(md::ModelDef, comp_def::ComponentDef, comp_name::Symbol;
     time_index = dim_keys(md, :time)
 
     if first == nothing
-        first = time_index[1]
-    elseif first < time_index[1]
+        first = 0
+    elseif first != 0 && first < time_index[1]
         error("Cannot add component $name with first time before first of model's time index range.")
```
rjplevin: As above.

test/test_components.jl
```diff
@@ -99,4 +96,76 @@ reset_compdefs()
...
+ci = m.mi.components[:C]    # Get the component instance
+@test ci.first == 2010      # The component instance's first and last values are the same as in the comp def
+@test ci.last == 2090
+
+set_dimension!(m, :time, 2000:2200) # Reset the time dimension
+cd = m.md.comp_defs[:C]     # Get the component definition in the model
+@test cd.first == 2010      # First and last values should still be the same
+@test cd.last == 2090
+
+Mimi.build(m)               # Build the model
```
rjplevin: Might as well use the new function to increase test coverage!
`@test first_period(md, cd) == 2010`

rjplevin: Ah, I see you use it below.

src/core/connections.jl
```diff
@@ -121,7 +123,7 @@ function connect_param!(md::ModelDef,
         dst_dims  = dimensions(dst_param)
 
         backup = convert(Array{number_type(md)}, backup) # converts number type and, if it's a NamedArray, it's converted to Array
-        first = first_period(dst_comp_def)
+        first = first_period(dst_comp_def, md)
```
rjplevin: Arg order needs to be flipped.

Issue #321, opened 2018-09-18
How to represent data in components that have different first/last times?


rjplevin:
As I'm migrating the code to 1.0, I'm encountering errors in tests that are sometimes unclear how to fix. Here's one...

We allow components to have a different first and last date than the dimension in question would suggest. So if `:time` is, say, 2010-2100, a component might start in, 2050. It's `run_timestep` won't be called for years 2010-2049.

Currently,  the component param/var allocates storage for the whole dimension, with the value for the variable's first year at index 1. The vector is padded with `NaN`s to the length of the dimension. I'm not sure how conscious a decision this was, but this is how it works now.

The `getdataframe` function handles it differently: values for the years supported by the component are returned, but these are padded before / after (along the time dimension) for any years that component excludes. 

Pictorally, where the parameter looks like this:
```
x x x x x x x x x x x NaN NaN NaN
^ first             ^last
```
a column in the dataframe might look like this:
```
NaN NaN x x x x x x x x x x x NaN
        ^ first             ^ last
```
It makes sense to use the latter format to store vars and params since this facilitates combining data across components with different first/last periods. I'm inclined to change the storage in components to work this way, but I wanted to get input from you guys first.

lrennels:
I've actually been wondering about this one too, and had the same question about whether it was consciously done.  I think changing the storage in components to work like the latter makes a lot more sense.

ckingdon95:
I think we should have the allocation make the parameter array the length of the dimension in the component, without the trailing NaNs. This would solve the tests that are failing in test_connectorcomp, and allow the same "fast indexing" that was the original intention (where the array and the index have the same `first` value). I'm happy to try this today and see what happens, if we think it's worthwhile to try

ckingdon95:
I think it makes sens for `getdataframe` to display the full length array with NaN's, so there's no ambiguity about if you're looking at the model's time dimension or the component's time dimension, but what would we gain by having the internal storage do this as well? I think it's more clear (and potentially faster) if the internal storage only has an array for the needed length.

rjplevin:
Actually, it should pad with `missing` values since `NaN` generally means an arithmetic operation failed, while what we really have is a "no data" situation.

ckingdon95:
Hmm actually your solution wouldn't be causing more integer division, since every array would use the same index, so maybe it's not slower...

lrennels:
Oh I was asking you both!

ckingdon95:
[...]

PR #331, opened 2018-10-11
WIP Datum Storage

ckingdon95:
This changes the array allocation for datum in components whose time dimensions are shorter than the model's time dimension. Now, TimestepArrays are allocated for the full length of the model's time index, and any cells before or after the component's first or last timestep values are padded with `missing` values. Addresses issue #321 

To do/questions:

- [x] Need to make final decision on if we pad with `missing`, `nothing`, or `NaN`
- [x] Do we want `getdataframe` to use `missing` or `nothing` instead of NaNs as well?
- [ ] I made a new test file "test_datum_storage.jl" to explicitly show the stuff we just changed, but these could probably get deleted or moved into the "test_connectorcomp.jl" test file. They are redundant, but they explicitly show the functionality, so not sure if we should keep them or not
- [x]  If backup data (during connections between shorter and longer components) has more than one dimension, only the size of the time dimension is checked against its size. Should we check other dimensions as well? How do we do this if the other dimensions aren't named/aren't necessarily in the same order as the model's dimensions?
- [x] should the provided backup data be required to be the full length of the model's time index, or just the full length of the longer component that it's filling in for?
- [x] When the `connect_param!` function is used with backup data, a ConnectorComp is added under the hood during build. This component does not show up in the model definition's list of components, but it shows up in the model instance's list of components. Is that okay?
- [ ] should I delete the `_vars_NT_type` function in "build.jl" that seems to be deprecated? (it's not called anywhere)
- [x] need to add tests for variable timesteparrays (to increase diff coverage)
- [x] need to add tests for error case (backup data wrong size)

### DIFF
--- a/src/components/connector.jl
+++ b/src/components/connector.jl
@@ -6,12 +6,14 @@ using Mimi
     output =  Variable(index = [time])
 
     function run_timestep(p, v, d, ts)
-        if !ismissing(p.input1[ts])
+        try 
             v.output[ts] = p.input1[ts]
-        elseif !ismissing(p.input2[ts])
-            v.output[ts] = p.input2[ts]
-        else
-            error("Neither of the inputs to ConnectorCompVector have data for the current timestep: $(gettime(ts)).")
+        catch 
+            try 
+                v.output[ts] = p.input2[ts]
+            catch 
+                error("Neither of the inputs to ConnectorCompVector have data for the current timestep: $(gettime(ts)).")
+            end
         end
     end
 end
@@ -25,12 +27,14 @@ end
 
     function run_timestep(p, v, d, ts)
         for r in d.regions
-            if !ismissing(p.input1[ts, r])
+            try 
                 v.output[ts, r] = p.input1[ts, r]
-            elseif !ismissing(p.input2[ts, r])
-                v.output[ts, r] = p.input2[ts, r]
-            else
-                error("Neither of the inputs to ConnectorCompMatrix have data for the current timestep: $(gettime(ts)).")
+            catch 
+                try 
+                    v.output[ts, r] = p.input2[ts, r]
+                catch 
+                    error("Neither of the inputs to ConnectorCompMatrix have data for the current timestep: $(gettime(ts)).")
+                end
             end
         end
     end
--- a/src/components/connector.jl
+++ b/src/components/connector.jl
@@ -6,9 +6,9 @@ using Mimi
     output =  Variable(index = [time])
 
     function run_timestep(p, v, d, ts)
-        if hasvalue(p.input1, ts)
+        if !ismissing(p.input1[ts])
             v.output[ts] = p.input1[ts]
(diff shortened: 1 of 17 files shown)


src/components/connector.jl
```diff
@@ -6,12 +6,14 @@ using Mimi
...
 
     function run_timestep(p, v, d, ts)
-        if hasvalue(p.input1, ts)
+        try 
             v.output[ts] = p.input1[ts]
-        elseif hasvalue(p.input2, ts)
-            v.output[ts] = p.input2[ts]
-        else
-            error("Neither of the inputs to ConnectorCompVector have data for the current timestep: $(gettime(ts)).")
+        catch
```
ckingdon95: @rjplevin @davidanthoff Now that I've added the new error to `getindex(timesteparray, timestep)` in the case of missing values, ConnectorComp relies on try/catch statements to determine if the shorter data is missing and if it should instead use the backup data. Is there a better way to make sure it's only catching that specific error, besides comparing the string of the error message? 

Alternatively, instead of relying on try/catch statements, I could add new parameters to the @defcomp that indicate the first and/or last years to use the shorter data for. 

```
@defcomp ConnectorCompVector begin
    input1 = Parameter(index = [time])
    input2 = Parameter(index = [time])
    output =  Variable(index = [time])
    short_first = Parameter()
    short_last = Parameter()

    function run_timestep(p, v, d, ts)
        if gettime(ts) >= short_first && gettime(ts) <= short_last
            v.output[ts] = p.input1[ts]
        else
            v.output[ts] = p.input2[ts]
        end
    end
end
```
These parameter values would be set when the connectorcomps are added during build, based on the `first` and `last` values of the source component. 

Which implementation would you prefer?

test/test_connectorcomp.jl
```diff
@@ -35,109 +21,146 @@ end
...
 
-add_comp!(m, ShortComponent; first=2100)
-add_comp!(m, Mimi.ConnectorCompVector, :MyConnector) # Rename component in this model
-add_comp!(m, LongComponent; first=2000)
+#------------------------------------------------------------------------------
+#  1. Manual way of using ConnectorComp
+#       TODO: are we no longer allowing users to add ConnectorComps manually?
+#       Currently deprecated because of checks for full/backup data during
+#       calls to connect_param!
+#------------------------------------------------------------------------------
```
ckingdon95: @davidanthoff 
For now I've commented out this test for the case of manually adding a ConnectorComp to a model. I added the error to `connect_param!` that will throw an error if a connection is made from a longer component to a shorter component without providing backup data. This therefore doesn't allow a user to add a ConnectorComp explicitly, because then if they call `connect_param(m, :ConnectorComp, :input, :Short, :var)` the new error will be thrown. I can add a work around for this so that the error is not thrown if the component is a ConnectorComp, or we can just choose to not support the manual use of ConnectorComps. Which do you prefer?

PR #340, opened 2018-10-23
Offset

lrennels:
This step removes the use of the `offset` keyword from the v0.5.0 API.  Once approved, this should be tagged as v0.5.1, although this does diverge from the current `master` branch ... @davidanthoff I could use your help for this merge and tag to make sure it goes smoothly.

Question:  Do we want to let `plot_comp_graph` plot cycles (as it does now) or remove this plotting feature for now?

### DIFF
--- a/docs/src/integrationguide.md
+++ b/docs/src/integrationguide.md
@@ -58,11 +58,6 @@ In an effort to standardize the function naming protocol within Mimi, and to str
 
 Changes to various optional keyword arguments:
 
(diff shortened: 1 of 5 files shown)


davidanthoff:
I've closed this PR for now and pushed the same commits on a branch ``release-0.5``, and will tag in a second. I think for now we can leave the plotting as is, given that v0.5.x is really a pretty temporary release.

PR #346, opened 2018-10-30
Anonymous dimensions

rjplevin:
I've restored support for dimensions given simply as an integer, and all the tests pass, but I haven't tested it on a "real" model that uses this feature. 

@lrennels, would you be willing to test this on the model that illuminated the problem? If that works, and all other tests pass, I think we're good to go. (DONE, works on fund)

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -13,6 +13,7 @@ FileIO
 GraphPlot 0.3.0
 JSON
 LightGraphs
+Logging
 MacroTools
 MetaGraphs
 NamedArrays
--- a/src/core/build.jl
+++ b/src/core/build.jl
@@ -56,18 +56,6 @@ function _instantiate_datum(md::ModelDef, def::DatumDef)
     return value
 end
 
-# Deprecated?
-# function _vars_NT_type(md::ModelDef, comp_def::ComponentDef)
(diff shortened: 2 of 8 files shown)


lrennels:
@rjplevin I'm happy to test this on `fund`, however right now we don't have `fund` up to Julia v1.0, only Julia v0.6 ... are these changes that could be applied to Mimi v0.5.1 so that I can test it there?

rjplevin:
I don't know if it will work on v0.5.1. How close is FUND to being v1.0-ready?

lrennels:
It's pretty close, @davidanthoff needs to check the marginal damages work and request changes, then we can merge `integration` and I can do the `v1.0` step, which has tended to be quite fast.

lrennels:
So one possible work flow could be:

1. tag the `v0.5.2` patch which is really just docs and a more explicit error to help people who run into the issue with anonymous dimensions
2. as soon as we get `fund` up to `v1.0`, test your branch with that and then merge (or merge your branch fist and test with `fund` afterwards if we feel pretty confident)

davidanthoff:
I would actually suggest that we just merge this to ``master``, so that this fix will only become available for the version that runs on julia 1.0.

PR #353, opened 2018-11-15
Change filepath of window in explore call

lrennels:


davidanthoff:
You'll also need to add a ``using FilePaths`` somewhere.

PR #359, opened 2018-11-29
WIP Tutorials

lrennels:
This PR attempts to build out four tutorials for users, and is currently a work in progress though would benefit from ongoing comments/advice from the team. I'd like to get a first working version out rather quickly in the next two weeks or so, and then we can always add and improve it.  I think this is a useful addition for new users and thus it would be smart to fast-track at least a first version.

@ckingdon95 @davidanthoff @rjplevin I'm going to keep working on this today, editing etc. since it's a rough draft, but it's ready for feedback/tearing apart because I'd like to get a first version out soon just so we have a working version I can point users to


docs/src/tutorials/tutorial_1.md
```diff
@@ -0,0 +1,127 @@
...
+The first step in this process is downloading the FUND model.  Open a Julia REPL (here done with the alias command `Julia`) and navigate to the folder where you would like to download FUND.
+
+```
+Julia 
+cd("directory")
+```
+
+Next, clone the FUND repository from Github, enter the repository, and checkout the `next` branch which holds the Julia version of FUND.
+```
+git clone https://github.com/fund-model/fund.git
```
davidanthoff: This doesn't work in the julia REPL, right? Do we need some sort of indication what to execute in the REPL, and what somewhere else?

docs/src/tutorials/tutorial_1.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Step 2. Run FUND
+
+The next step is to run FUND.  If you wish to first get more aquainted with the model itself, take a look at the provided online documentation.  
+
+In order to run FUND, you will need to navigate to the source code folder, labeled `src`, and run the main fund file `fund.jl`.  This file defines a new [module](https://docs.julialang.org/en/v1/manual/modules/index.html) called `Fund`, which exports the function `getfund`, a function that returns a version of fund allowing for different user specifications.  Note that in order to allow access to the module, we must call `using .Fund`, where `.Fund` is a shortcut for `Main.Fund`, since the `Fund` module is nested inside the `Main` module. After creating the model `m`, simply run the model using the `run` function.
+
+```
+include("src\fund.jl")
+using .Fund
```
davidanthoff: Do we need the dot, actually? I'm pretty sure not :)

docs/src/tutorials/tutorial_1.md
```diff
@@ -0,0 +1,127 @@
...
+Note that these steps should be relatively consistent across models, where a repository for `ModelX` should contain a primary file `ModelX.jl` which exports, at minimum, a function named something like `getModelX` or `construct_ModelX` which returns a version of the model, and can allow for model customization within the call.
+
+In this case, the function `getfund` has the signature
+``` 
+getfund(; nsteps = default_nsteps, datadir = default_datadir, params = default_params)
+```
+Thus there are no required arguments, although the user can input `nsteps` to define the number of timesteps (years in this case) the model runs for, `datadir` to define the location of the input data, and `params`, a dictionary definining the parameters of the model.  For example, if you wish to see only the first 100 timesteps,you may use
+```
+include("src\fund.jl")
+using .Fund
```
rjplevin: Use Unix slashes "/" since these work on Windows, Mac, and Linux. Backslashes work only on Windows.

docs/src/tutorials/tutorial_2.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Parametric Modifications: DICE Example
+
+### Step 1. Download DICE
+
+The first step in this process is downloading the DICE model.  Open a Julia REPL (here done with the alias command `Julia`) and navigate to the folder where you would like to download DICE.
+```
+Julia 
+cd("directory")
+```
```
rjplevin: Add a comment after `cd` command indicating that user must provide real dir name

docs/src/tutorials/tutorial_2.md
```diff
@@ -0,0 +1,125 @@
...
+
+In this case, the function `construct_dice` has the signature
+``` 
+construct_dice(params=nothing)
+```
+Thus there are no required arguments, although the user can input `params`, a dictionary definining the parameters of the model. 
+
+### Step 3. Altering Parameters
+
+In the case that you wish to alter an exogenous parameter, you may use the `update_param!` function.  For example, in DICE the parameter `fco22x` is the forcings of equilibrium CO2 doubling (Wm-2), and exists in the components `climatedynamics` and `radiativeforcing`.  If you wanted to change this value from its default value of `3.200` to `3.000` in both components,you would use the following code:
```
rjplevin: Since W per square meter doesn't render nicely, you might want to spell it out rather than writing "Wm-2" which may be confusing.

docs/src/tutorials/tutorial_2.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+Note that here we use the `update_timesteps` flag and set it to `true`, because since we have changed the time index we want the time labels on the parameters to change, not simply their values.
+
+## Component and Structural Modifications: The API
+
+Most model modifications will include not only parametric updates, but also strutural changes and component modification, addition, replacement, and deletion along with the required re-wiring of parameters etc. The most useful functions of the common API, in these cases are likely **`replace_comp!`, `add_comp!`** along with **`Mimi.delete!`** and the requisite functions for parameter setting and connecting.  For detail on the public API functions look at the API reference [here](http://anthofflab.berkeley.edu/Mimi.jl/dev/reference/). 
+
+Users who wish to modify the component structure would also do well to look into the **built-in helper components`adder`, `ConnectorCompVector`, and `ConnectorCompMatrix`** in the `src\components` folder, as these can prove quite useful.  
+
```
rjplevin: src/components

docs/src/tutorials/tutorial_1.md
lrennels: If I don't include the dot, I get:

```
ERROR: ArgumentError: Package Fund not found in current path:
- Run `Pkg.add("Fund")` to install the Fund package.
```

PR #361, opened 2018-11-29
Fix syntax in docs

lrennels:


docs/src/tutorial.md
```diff
@@ -7,7 +7,7 @@ In this example, we will construct a stylized model of the global economy and it
 There are two main steps to creating a component, both within the  `@defcomp` macro which defines a component:
 
 * List the parameters and variables.
-* Use the run_timestep function `run_timestep(p, v, d, t)` to set the equations of that component.
+* Use the `run_timestep` function `run_timestep(p, v, d, t)` to set the equations of that component.
```
rjplevin: Simpler:

Use the function `run_timestep(p, v, d, t)` to set...

davidanthoff:
I just merged for now, otherwise we can't really see how these updates look in the docs. If there is more todo, just open another PR :)

PR #379, opened 2018-12-19
Update dependency paths

### LINKED ISSUE #377: CI travis fails
The now included dependency tests fail

lrennels:
Fixes #377 
- need to update `fund` path once the `1.0` branch is merged

### DIFF
--- a/test/test_dependencies.jl
+++ b/test/test_dependencies.jl
@@ -52,11 +52,8 @@ function run_dependency_tests(dependencies=dependencies)
 
         #test the package
         try
-            println("trying to run $package_name tests")
             run(`$(Sys.BINDIR)/julia $process`)
-            println("ran $package_name tests")
         catch e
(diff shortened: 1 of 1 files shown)


lrennels:
@davidanthoff it's not quite clear to me why RICE is erring here, and FUND is also erring soon since it's not even getting through it's tests, one of which (marginal damages) I expect to fail.  

I was stepping through the `test_dependencies.jl` code to try and figure out what was going on, and although I can't run `run(`$JULIA_HOME/julia $process`)` from the REPL (I assume that's a Travis thing?), `include(runtests.jl)` works fine from there so I'm a little stuck on what's going on.  Any ideas? I thought maybe it's an `add Pkg` problem but I think I got them all.  I'll keep digging.  

For now I could also remove this so that the docs rebuild, which we want to happen to fix some bugs, but I'd rather just solve the issue!

PR #382, opened 2019-01-18
Make locally built docs work

rjplevin:
Set environment var JULIA_NO_LOCAL_PRETTY_URLS to anything to engage.


davidanthoff:
But the doc build fails, so we should probably fix that first.

rjplevin:
Sheesh. So now travis passed but the nightly appveyor run timed out with "Build execution time has reached the maximum allowed time for your plan (60 minutes)." Spurious? Fixable?

PR #385, opened 2019-01-29
Update module use in tutorials; fix syntax bug

lrennels:


docs/src/tutorials/tutorial_3.md
```diff
@@ -89,28 +89,33 @@ module my_model
...
+function construct_model()
+	m = Model()
 
-set_dimension!(m, :time, collect(2015:5:2110))
+	set_dimension!(m, :time, collect(2015:5:2110))
 
-# Order matters here. If the emissions component were defined first, the model would not run.
-add_comp!(my_model, grosseconomy)  
-add_comp!(my_model, emissions)
+	# Order matters here. If the emissions component were defined first, the model would not run.
```
rjplevin: Shouldn't these be `m` rather than `my_model`?

lrennels: @rjplevin yes, I didn't do this carefully enough!  I just pushed another set of changes that more comprehensively update Tutorial 3 and also change the actual `examples/tutorial` code to match and not use `@defmodel`

PR #391, opened 2019-02-03
Mcs bugfix

lrennels:
This PR fixes the MCS bug that did not increment `models_to_run` when a model, or vector of models, were passed to `run_mcs`.  It also augments the set of `run_mcs` methods to be comprehensive about `trials::Int` and `trials::Union{Int, Vector{Int}, AbstractRange{Int}}`.

### DIFF
--- a/src/mcs/montecarlo.jl
+++ b/src/mcs/montecarlo.jl
@@ -492,7 +492,7 @@ function run_mcs(mcs::MonteCarloSimulation, m::Model, trials::Union{Vector{Int},
     return run_mcs(mcs, trials, models_to_run; kwargs...)
 end
 
-function run_mcs(mcs::MonteCarloSimulation, m::Model, trials=mcs.trials, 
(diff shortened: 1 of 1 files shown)


lrennels:
@rjplevin I agree that we should keep a capability to run selected models from the list.  A few thoughts though ...

(1) and edge case

I call `run_sim(sim, m, trials = N, models_to_run = 1)` on a sim that already has `length(sim.models) = 3`.  If we use the `models_to_run::Int = length(sim.models) +1`, now I’ve added a fourth model to the list, and it’s going to run models 1 and 2.

(2) Right now we use the following, which wouldn't allow you to run, say models 2 and 3.  I think for now that's fine though.
```
for m in mcs.models[1:models_to_run]    # note that list of models may be changed in scenario_func
                    @debug "Running model"
                    run(m, ntimesteps=ntimesteps)
                end
```

PR #392, opened 2019-02-12
Add @modelegate for dim functions

lrennels:


lrennels:
@jrising I think this well take out some verbosity of you needing to call `ModelDef(dim_count ...)` for various dimension focused functions.  Does this help solve the problem you were asking about?  Feel free to say no so we can make sure we're doing what you need!

jrising:
Great, thank you @lrennels !

PR #398, opened 2019-02-15
Separate set_models! from run_mcs

rjplevin:
Partial fix for `run_mcs` issue. @lrennels should take a look...

### DIFF
--- a/docs/src/internals/montecarlo.md
+++ b/docs/src/internals/montecarlo.md
@@ -2,13 +2,15 @@
 
 ## Overview
 
-Monte Carlo Simulation support consists of three primary user-facing elements:
+Monte Carlo Simulation support consists of four primary user-facing elements:
 
 1. The `@defmcs` macro, which defines random variables (RVs) which are assigned distributions and associated with model parameters, and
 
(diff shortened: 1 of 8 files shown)


lrennels:
@rjplevin per our conversation we need a bit of design discussion for this.  Our incrementing idea doesn't quite hold because calling `run_mcs(mcs, m, ...)` multiple times keeps incrementing the `models_to_run` without checking if `m` is an element of `models_to_run`.  One solution would be to separate steps so that a user has to (1) `set_model(s)!` and (2) `run_mcs`.  That would allow `run_mcs` to take the argument `models_to_run` indicating the **indices** of the models the user wants to run without the complication of calling `set_model!` within `run_mcs`.  Thoughts?

rjplevin:
I agree. This simplifies the API and avoids hard-to-explain workarounds. Will you do this? (Thanks!)

lrennels:
@rjplevin sure I'll do it, I might also remove `set_model` and just use`set_models!` function with two methods, one for a single model and one for a vector of models.

lrennels:
Similarly, should `models_to_run` be the first n models, or should we allow users to pick which models?  Just throwing out design decisions that seem like they need to be wrapped up!

Issue #390, opened 2019-02-02
restrict the visualization to a given time period with explorer?


hbenveniste:
It would be great to be able to chose a specific time period when visualizing results for explorer. For instance focusing on 1950-2100 or 1950-2300, rather than all the way to year 3000.

davidanthoff:
Maybe we could essentially do [this](https://discourse.julialang.org/t/interactive-plot-acting-as-a-range-slider-tool-for-another-plot/20392/4), and add an option to configure a default time range for a given model somehow?

lrennels:
@davidanthoff good idea I'll look into this!

davidanthoff:
First version: add keyword args to `explore`.

PR #400, opened 2019-02-16
Explorer timewindow

lrennels:
This PR is a response to Issue #390.

The main contributions are:

- For single and multiline plots in `explorer`, allow the user to interactively select a time window of interest from a sub-plot showing the full range.
- export `save`

Current TODO:

- It's not clear whether having a `save_vega_plot` function is a good idea.  My original thought was that it prevents users from having to include `using VegaLite` before calling `save("filepath", plot)`.  One user had trouble with this on the tutorials.  The function name could just be `save`, but I'm not sure what the protocol is for adding methods to `FileIO`.  @davidanthoff @rjplevin thoughts on this extra function? - **SOLVED: Just use `export save` and don't need `save_vega_plot`**

- Right now saving a line plot fails, because `save` doesn't work for interactive plots.  **SOLVED: warning appears but plot is still saved accurately**


davidanthoff:
Aweseome!

I don't think we need `save_vega_plot`. I believe you can just add `export save` somewhere, and then the `FileIO.save` function will be exported by Mimi as well. I think that would solve the issue, right?

lrennels:
@davidanthoff good point I'll just export `save`, I think we're set now.  Using `save` on an interactive plot prints these lines:
```
WARN Can not resolve event source: window
WARN Can not resolve event source: window
```
But saves fine, so I think we're good for now on that.

davidanthoff:
Ah, yes! If you try to save an interactive plot in a non-interactive file format, that message shows up. But you can also save a plot as a `*.vegalite` file, and then if you open it in Jupyter Lab all the interactive aspects are preserved :)

lrennels:
Oh interesting!  I'll add that note to the docs now, then I think we can merge.

src/explorer/explore.jl
```diff
@@ -65,3 +67,12 @@ function explore(m::Model, comp_name::Symbol, datum_name::Symbol)
...
 end
+
+"""
+    function vega_plot(filepath::String, p::VegaLite.VLSpec{:plot})
+
+Save a the plot `p` as `filepath`.  Simple wrapper function for VegaLite's `save` function.
+"""
+function save_vega_plot(filepath::String, p::VegaLite.VLSpec{:plot})
+    VegaLite.save(filepath, p);
+end
```
davidanthoff: I think we can remove this whole function before we merge, right?

lrennels: Yes, thought I did that, thanks!

Issue #402, opened 2019-02-19
Scale y axis in explorer based on time window that is selected


davidanthoff:
For example, here:

![image](https://user-images.githubusercontent.com/1036561/52985088-0c464e80-33a7-11e9-8961-be7f9f3ad9c3.png)

it would be nice if the y axis in the top plot had a smaller range so that one can make sense of the lines :)

This might be slightly tricky, because I guess one needs to adjust the scale somehow based on the selection, right?

lrennels:
Good idea, I'm surprised that this isn't done in the examples.  The x axis domain scale is pulled from the brush selection, but simply doing the same with they y axis isn't effective, perhaps because the brush selection is only done on the x encoding.  I'll look into it.

PR #403, opened 2019-02-19
Add y axis to brush selection

lrennels:
@davidanthoff this addresses #402 by including the y axis in the brush selection.  If we wanted to just brush the x axis and have the y axis scale automatically, there may be a way to filter the data as we go and then change the y axis based on the new selection.  VegaLite warns that this is less efficient, since it involves constantly evaluating the data based on the changing conditions as opposed to just using the same data and a new scale, but I think it would be doable.  This might be preferred if it is too tedious to select for clumped-together, low y values.


davidanthoff:
Yeah, I think it would be nicer if the selection was only for the x axis, and then the height of the y axis automatically adjusted based on the min and max value in the selected x axis region. Maybe we can try whether that has performance issues, and only if it does, go with this PR here?

lrennels:
Sure, for my reference see Selection Predicate here: https://vega.github.io/vega-lite/docs/filter.html

lrennels:
@davidanthoff Alright this should do the trick.  Note that in this version, you must brush over at least two year intervals (so ten years here since we have 5 year intervals) for anything to show up.  It seems that VegaLite is off-by-one on the x axis, showing labels 2014 to 2109 for the example instead of 2015 to 2110.  I'll work on figuring that out.

lrennels:
off-by-one error solved using `"timeUnit" => "utcyear"` instead of `"timeUnit" => "year"` see https://github.com/vega/vega-lite/issues/4044

PR #405, opened 2019-02-21
added note on package installation

fhoces:
This could be useful for users starting from a fresh installation of Julia (like me!)

### DIFF
--- a/docs/src/tutorials/tutorial_1.md
+++ b/docs/src/tutorials/tutorial_1.md
@@ -38,7 +38,7 @@ Julia
 cd(<fund-directory-path>) # fund-directory-path is a placeholder for the string describing your the file path of the downloaded `fund` folder from Step 1.
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/tutorial_1.md
```diff
@@ -38,7 +38,7 @@ Julia
...
 ```
 
-Next, run the main fund file `fund.jl`.  This file defines a new [module](https://docs.julialang.org/en/v1/manual/modules/index.html) called `Fund`, which exports the function `getfund`, a function that returns a version of fund allowing for different user specifications.  Note that in order to allow access to the module, we must call `using .Fund`, where `.Fund` is a shortcut for `Main.Fund`, since the `Fund` module is nested inside the `Main` module. After creating the model `m`, simply run the model using the `run` function.
+Next, run the main fund file `fund.jl` (make sure to have the packages `Distributions` and `StatsBase` installed -using `Pkg.add("StatsBase")`-.  This file defines a new [module](https://docs.julialang.org/en/v1/manual/modules/index.html) called `Fund`, which exports the function `getfund`, a function that returns a version of fund allowing for different user specifications.  Note that in order to allow access to the module, we must call `using .Fund`, where `.Fund` is a shortcut for `Main.Fund`, since the `Fund` module is nested inside the `Main` module. After creating the model `m`, simply run the model using the `run` function.
```
lrennels: I would recommend that instead of instructing users to using the `Pkg` package, thus needing to type `using Pkg` and then use `Pkg.xxx`, we should instruct users to use the Pkg mode in the Julia REPL, which is accessed by typing `]` and exited by pressing delete.  For example here you would instruct:

First enter the Julia package mode by typing `]`, then add the necessary packages with `add Distributions` and `add StatsBase`.

PR #409, opened 2019-02-28
- Added support for distributions that produce a matrix of values

rjplevin:
Added support for distributions that produce a matrix of values on each draw, as requested by @hbenveniste on the Mimi forum. See `tests/mcs/test_reshaping.jl` and the modified 2-region model in the `tests/mcs/test-model` directory for a simple example.

We might wait until the user tests this branch before merging, although all tests pass, so the changes appear to do no harm.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
   - linux
   - osx
 julia:
-  - 1.0
   - 1.1
   - nightly
 matrix:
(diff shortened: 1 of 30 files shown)


rjplevin:
I think we should merge this now. The changes are minor, but one to note is the REQUIRE file change to use julia 1.1. Are we ok with that? (I hope so!)

davidanthoff:
I think @lrennels  should probably review this and then you guys can merge?

PR #415, opened 2019-03-05
Eliminate global comp dict

rjplevin:
- Eliminated global dict of component defs
- Fixed uses of eval to avoid precompilation issues

### DIFF
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -87,7 +87,7 @@ end
 
 # Components are defined here to allow pre-compilation to work
 function __init__()
-    compdir = joinpath(dirname(@__FILE__), "components")
+    compdir = joinpath(@__DIR__, "components")
     load_comps(compdir)
 end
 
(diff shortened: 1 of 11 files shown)


src/Mimi.jl
```diff
@@ -87,7 +87,7 @@ end
 
 # Components are defined here to allow pre-compilation to work
 function __init__()
-    compdir = joinpath(dirname(@__FILE__), "components")
+    compdir = joinpath(@__DIR__, "components")
     load_comps(compdir)
```
davidanthoff: Do we still need to do this in `__init__``? With these fixes, shouldn't we just be able to include these files in the normal way now? We don't have to change this in this PR, just wondering...

rjplevin: Agreed. I quickly tried to change this but MimiDice2013 complained about loading Mimi from cache. Rather than debug this now, I'll merge what we have.

PR #434, opened 2019-03-14
WIP Allow any placement of time dimension

ckingdon95:
Description of changes:
- TimestepArrays now have an additional type parameter, `ti` which is an integer value equal to the time index position of that array. (I'm not particularly attached to the name `ti`, we can pick something better if we want)
- additional `getindex` and `setindex` functions for `TimestepMatrix` with `ti` equal to 2 were added 
- `getindex` and `setindex` functions for `TimestepArrays` have all been changed, similar to the following example:
```
function Base.getindex(arr::TimestepArray{FixedTimestep{FIRST, STEP}, T, N, ti}, idxs::Union{FixedTimestep{FIRST, STEP, LAST}, AnyIndex}...) where {T, N, ti, FIRST, STEP, LAST}
	idxs1, ts, idxs2 = idxs[1:ti - 1], idxs[ti], idxs[ti + 1:end]
	return arr.data[idxs1..., ts.t, idxs2...]
end
```
- there's a helper function called `get_time_index_position` that gets used during build to get the appropriate `ti` value based on the parameter or variables dimension order.

Todo:
- I've written tests locally that test most of the functionality, but I want to place them in appropriate test files rather than adding an additional test file. 
- I tried to get explorer working (right now it displays values as a bar graph if time is the second dimension) but I haven't pushed those changes yet because I haven't got it to work
- we might want to implement some better error messaging if users mess up their dimension ordering


rjplevin:
> `idxs1, ts, idxs2 = idxs[1:ti - 1], idxs[ti], idxs[ti + 1:end]`

Personal taste, but I find this "3 assignments in one" formulation less readable than 3 simple assignments on separate lines.

Also, should the new parameter be `TI` in uppercase for consistency?

rjplevin:
For cases like these:
```julia
function next_timestep(ts::FixedTimestep{FIRST, STEP, LAST}) where {FIRST, STEP, LAST}
	if finished(ts)
	    error("Cannot get next timestep, this is last timestep.")
	end
	return FixedTimestep{FIRST, STEP, LAST}(ts.t + 1)
end

function next_timestep(ts::VariableTimestep{TIMES}) where {TIMES}
	if finished(ts)
	    error("Cannot get next timestep, this is last timestep.")
	end
	return VariableTimestep{TIMES}(ts.t + 1)		
end
```
Can't we have a single function:

```julia
function next_timestep(ts::T) where {T <: AbstractTimestep}
	if finished(ts)
	    error("Cannot get next timestep, this is last timestep.")
	end
	return T(ts.t + 1)		
end

**Comment by ckingdon95** [comment]
- generally I would agree that the triple assignment reads worse, but in this case it appears so many times in the "time.jl" file that I personally prefer the brevity. (an alternative would be to define a helper function like `split_index(idxs, ti)` that returns all three that we could use everywhere. thoughts?)
- happy to go with capital `TI`
- you're definitely right about the pairs of functions, although that's not completely related to this PR. but I could make those changes here now if we want (since it applies to several of the functions modified here)

**Comment by ckingdon95** [comment]
also just a note on the innards of "time.jl": some of the functions for `VariableTimesteps` that looked like this:
```
function Base.getindex(mat::TimestepMatrix{VariableTimestep{D_FIRST}, T, 1}, ts::VariableTimestep{T_FIRST}, idx::AnyIndex) where {T, D_FIRST, T_FIRST}
	t = ts.t + findfirst(isequal(T_FIRST[1]), D_FIRST) - 1
	data = mat.data[t, idx]
	_missing_data_check(data)
end
```
I changed to this:
```
function Base.getindex(mat::TimestepMatrix{VariableTimestep{D_TIMES}, T, 1}, ts::VariableTimestep{T_TIMES}, idx::AnyIndex) where {T, D_TIMES, T_TIMES}
	t = ts.t + findfirst(isequal(T_TIMES[1]), D_TIMES) - 1
	data = mat.data[t, idx]
	_missing_data_check(data)
end
```
because I believe the parameter names `D_FIRST` and `T_FIRST` were misleading and that `D_TIMES` and `T_TIMES` are more appropriate. This doesn't change any functionality, I just noticed this while I was going through all of these functions.

rjplevin:
Defining a short function `split_index(idxs, ti)` sounds like a good approach. (Short funcs get inlined anyway, so there's no penalty.) It would be both readable and concise.

I'm agnostic about whether we consolidate the pairs of functions in this PR or in a separate one.

rjplevin:
For some reason, I'm not seeing an "Approve" button, but I think you can merge this if tests are all passing. Nice job!

ckingdon95:
Okay I'll make those few changes but I'm not going to merge it until I add more tests! and get explorer working (maybe I can elicit @lrennels for that)

lrennels:
@ckingdon95 happy to help if things are confusing in `explorer`. It assumes that if there is a `time` dimension, and thus it needs to pick a `line` or `multiline` plot, that `time` is the first dimension. The easiest way to fix it is probably to expand that check in `_spec_for_item` on line 33 of `buildspecs.jl` and then rearrange the dataframe that is passed to `createspec_multiline` or `createspec_line`.  Happy to continue to discuss!

ckingdon95:
@lrennels I merged master back into this branch so the tests just re-ran. It looks like some of the dependency tests failed on Travis, but not for all of the deployments. You're more familiar with these tests than I am, would you mind taking a look and seeing if you think this is caused by something on this branch, or if this is just an unrelated Travis problem? Thanks!!

lrennels:
@ckingdon95 we're getting this exact same error in the current `Mimi` build, so I don't think there's an issue with this branch!  I will look into what's going on with the Mimi build but it seems you could merge this and we'd be in the same place we are already :)

PR #440, opened 2019-03-25
Handle doc string before run_timestep in @defcomp.

rjplevin:
As per James' posted issue.


rjplevin:
The build errors seem to be from Cairo, so I'm ignoring them for now...

Issue #444, opened 2019-03-29
Change filename option in generate_trials


ckingdon95:
Currently generate_trials! looks like this, where "" is the default filename instead of nothing:
```
function generate_trials!(sim::Simulation{T}, samplesize::Int;
                          filename::String="") where T <: AbstractSimulationData

    sample!(sim, samplesize)

    # TBD: If user asks for trial data to be saved, generate it up-front, or 
    # open a file that can be written to for each trialnum/scenario set?
    if filename != ""
        save_trial_inputs(sim, filename)
    end
end
```
Is there any reason we can't change it to this:
```
function generate_trials!(sim::Simulation{T}, samplesize::Int;
                          filename::Union{String, Nothing}=nothing) where T <: AbstractSimulationData
    sample!(sim, samplesize)
    if filename != nothing
        save_trial_inputs(sim, filename)
    end
end
```
@davidanthoff @rjplevin

rjplevin:
No reason. I thought it was simpler than declaring the Union, but it is more consistent to use `nothing` as the sentinel value.

lrennels:
#445

PR #445, opened 2019-03-30
Make generate_trials filename kwarg default to nothing

lrennels:
Change the `generate_trials` keyword argument `filename` default to `nothing` instead of `""`.  Addresses Issue #444.

### DIFF
--- a/src/mcs/montecarlo.jl
+++ b/src/mcs/montecarlo.jl
@@ -140,20 +140,19 @@ function get_trial(sim::Simulation, trialnum::Int)
 end
 
 """
-    generate_trials!(sim::Simulation{T}, samples::Int; filename::String="")
+    generate_trials!(sim::Simulation{T}, samples::Int; filename::Union{String, Nothing}=nothing)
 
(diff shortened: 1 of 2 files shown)


src/mcs/montecarlo.jl
```diff
@@ -140,20 +140,19 @@ function get_trial(sim::Simulation, trialnum::Int)
...
 function generate_trials!(sim::Simulation{T}, samplesize::Int;
-                          filename::String="") where T <: AbstractSimulationData
-
+                        filename::Union{String, Nothing}=nothing) where T <: AbstractSimulationData
     sample!(sim, samplesize)
 
     # TBD: If user asks for trial data to be saved, generate it up-front, or 
     # open a file that can be written to for each trialnum/scenario set?
-    if filename != ""
+    if filename != nothing
```
davidanthoff: VERY minor nitpick here: for comparisons with `nothing` (or `missing`), it is better to use `===` and `!==` because then the compiler can do some fancy things that speed stuff up. Entirely irrelevant here, just thought I'd mention it in general :)

PR #448, opened 2019-04-04
Add a various method for Empirical Distribution

lrennels:
@rjplevin these methods were added as I was trying to get `quantile.` to work, a can I may kick down the road but these methods are probably still worth adding.  Let me know if you think the three of them make sense.

### DIFF
--- a/src/mcs/EmpiricalDistribution.jl
+++ b/src/mcs/EmpiricalDistribution.jl
@@ -56,6 +56,10 @@ function Base.length(d::EmpiricalDistribution, args::Vararg{Integer,N}) where {N
     return length(d.values)
 end
 
+function Base.iterate(d::EmpiricalDistribution, args::Vararg{Integer,N}) where {N}
+    return iterate(d.values)
+end
+
 function Statistics.quantile(d::EmpiricalDistribution, args...)
     indices = quantile.(d.dist, args...)
     return d.values[indices]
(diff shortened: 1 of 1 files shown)


src/mcs/EmpiricalDistribution.jl
```diff
@@ -52,6 +52,18 @@ function Statistics.var(d::EmpiricalDistribution)
...
+    return length(d.values)
+end
+
+function Base.collect(d::EmpiricalDistribution, args::Vararg{Integer,N}) where {N}
+    return collect(d.values)
+end
+
+function Base.iterate(d::EmpiricalDistribution, args::Vararg{Integer,N}) where {N}
+    return iterate(d.values)
+end
```
davidanthoff: This strikes me as an incorrect implementation of the iteration interface :) I think you want:
```julia
Base.iterate(x::EmpiricalDistribution, s...) = iterate(x.x, s...)
```

src/mcs/EmpiricalDistribution.jl
```diff
@@ -52,6 +52,18 @@ function Statistics.var(d::EmpiricalDistribution)
     return var(d.values, d.weights, corrected=true)
 end
 
+function Base.length(d::EmpiricalDistribution, args::Vararg{Integer,N}) where {N}
+    return length(d.values)
+end
+
+function Base.collect(d::EmpiricalDistribution, args::Vararg{Integer,N}) where {N}
+    return collect(d.values)
+end
```
davidanthoff: I don't think we need this if we implement the iterator interface properly.

lrennels:
I think these updates are now a bit irrelevant since we went a different direction with the bug fix.  We can revive the conversation  later if needed.

PR #468, opened 2019-04-30
Remove some dependencies

davidanthoff:
This removes some unnecessary dependencies, and marks some of them as only needed for testing. Overall this should make installation of the package lighter weight and improve load times of the package.


src/mcs/montecarlo.jl
```diff
@@ -562,7 +562,7 @@ column_types(sim::Simulation{T}) where T <: AbstractSimulationData = [eltype(fld
 # Iteration support (which in turn supports the "save" method)
 #
 column_names(iter::SimIterator) = column_names(iter.sim)
-column_types(iter::SimIterator) = IterableTables.column_types(iter.sim)
+column_types(iter::SimIterator) = error("Not implemented") # Used to be `IterableTables.column_types(iter.sim)`
```
davidanthoff: @rjplevin the old code can not have worked because the `column_types` function was removed when we moved to julia 1.0. I assume this is just not used, given that no tests fail?

Issue #464, opened 2019-04-23
String and symbol parameters fail on run(m)


jrising:
If you define a parameter to have a type of String or Symbol, everything seems to work properly until you run the model, and then it gives an exception.

Here's a minimal example:
```
using Mimi

@defcomp Example begin
    param::String = Parameter()

    function run_timestep(p, v, d, t)
    end
end

m = Model()
set_dimension!(m, :time, 1:10)

add_comp!(m, Example)
set_param!(m, :Example, :param, "test")

run(m)
```

And the last line results in this error:
```
julia> run(m)
ERROR: MethodError: no method matching copy(::Symbol)
Closest candidates are:
  copy(::Expr) at expr.jl:36
  copy(::BitSet) at bitset.jl:46
  copy(::Markdown.MD) at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.0/Markdown/src/parse/parse.jl:30
  ...
Stacktrace:
 [1] copy(::Mimi.ScalarModelParameter{Symbol}) at /Users/jrising/.julia/dev/Mimi/src/core/defs.jl:735
 [2] (::getfield(Mimi, Symbol("##87#88")))(::Pair{Symbol,Mimi.ModelParameter}) at ./none:0
 [3] iterate at ./generator.jl:47 [inlined]
 [4] Dict{Symbol,Mimi.ModelParameter}(::Base.Generator{Dict{Symbol,Mimi.ModelParameter},getfield(Mimi, Symbol("##87#88"))}) at ./dict.jl:103
 [5] copy_external_params at /Users/jrising/.julia/dev/Mimi/src/core/defs.jl:731 [inlined]
 [6] copy(::Mimi.ModelDef) at /Users/jrising/.julia/dev/Mimi/src/core/defs.jl:771
 [7] build at /Users/jrising/.julia/dev/Mimi/src/core/build.jl:92 [inlined]
 [8] #run#113(::Int64, ::Nothing, ::Function, ::Model) at /Users/jrising/.julia/dev/Mimi/src/core/model.jl:395
 [9] run(::Model) at /Users/jrising/.julia/dev/Mimi/src/core/model.jl:390
 [10] top-level scope at none:0
```

Whereas this code works fine if I replace `String` with `Float64` and give a float parameter value.  The same kind of error happens with symbol parameters.

PR #469, opened 2019-05-01
Add method for copying params with string or symbol type

lrennels:
Addresses #464 @jrising

### DIFF
--- a/src/core/defs.jl
+++ b/src/core/defs.jl
@@ -732,6 +732,8 @@ end
 
 Base.copy(obj::ScalarModelParameter{T}) where T = ScalarModelParameter{T}(copy(obj.value))
 
+Base.copy(obj::ScalarModelParameter{T}) where T <: Union{Symbol, AbstractString} = ScalarModelParameter{T}(obj.value)
+
 Base.copy(obj::ArrayModelParameter{T})  where T = ArrayModelParameter{T}(copy(obj.values), obj.dimensions)
 
 function Base.copy(obj::TimestepVector{T_ts, T}) where {T_ts, T}


lrennels:
Yes, I tried 

```
using Mimi

@defcomp Example begin
    param::Array{String, 2} = Parameter()

    end
end

m = Model()
set_dimension!(m, :time, 1:10)

add_comp!(m, Example)
param = ["a" "b"; "c" "d"]

set_param!(m, :Example, :param, param)

run(m)
```
... and it works even without this PR so I think we're fine although I should probably also look into why that works!

jrising:
@lrennels Thanks for this!  But I'm surprised by the syntax for arrays (although I don't need arrays, so it's not pressing).  Should this be:
```
param::String = Parameter(index=[time])
```

I think it's better to be consistent on the way that Parameters are defined (that is, listing their indices) than to say that the `::` syntax always refers to the whole parameter variable.

lrennels:
I was thinking of this slightly differently, in that the `::` syntax would help us type check the parameter we bind to `param`, in which case it would not be a `String` but an `Array`. @davidanthoff @rjplevin what are your thoughts here?

lrennels:
So still my example was incorrect, but it would be 
```
param::param::Array{String, 2} = Parameter(index=[time])
```

rjplevin:
It's a choice between keeping it simple, in which case it would be:
`paramName::String = Parameter(index=[time])`
vs being somewhat more true to Julia syntax, which would be as Lisa suggested:
` param::Array{String, 1} = Parameter(index=[time])`

(I assume that `param::param::` was a typo, right?)

As a third alternative, we could avoid overloading Julia syntax and add keyword to our Parameter definition, e.g., 
`paramName = Parameter(index=[time], type=String)`

I think it would be less desirable to have to use full Julia syntax on the left side when the right side departs from this, and is somewhat redundant with the `index=` bit.

jrising:
My vote is for your third option, @rjplevin.  I've always felt that the param::Type syntax was a bit clunky.

davidanthoff:
The `paramName::MyType= Parameter(index=[time])` has been around for a long time and I'm pretty sure it is used in a number of places in models. So I think in the interest of not breaking things and not introducing another alternative syntax, I would stick with that.

rjplevin:
I appreciate the desire to avoid breaking things, but I'm not sure that now is the time to lock in choices that we might regret later on for not changing them when the user base was still very small.

We could support both the old style and the new, and recommend using the new syntax, then deprecate the old syntax at some point. A gradual off ramp...

lrennels:
Hmm I'm more inclined to stick with 
```Julia
param::Array{String, 1} = Parameter(index=[time])
```
because it seems more intuitive from a Julia perspective.  It's true that this might seem repetitive in some ways, but it's also not _necessary_ to parameterize `param` in this way so we aren't enforcing repetition.  To me it seems more inconsistent to use the first option
```Julia
param::String = Parameter(index=[time])
```
because it breaks from the intuition of type parameterization.

I can move this to a separate (open) issue for continued discussion, although it's not urgent.

Issue #474, opened 2019-05-08
Update parameter in Monte Carlo run


grantmcdermott:
The [package documentation](https://www.mimiframework.org/Mimi.jl/stable/tutorials/tutorial_2/) provides a nice example of how to update a parameters using the `!update_param` API method. However, near as I can tell, this method only works for the deterministic version of the model(s). I'm looking for a way to update parameters in the Monte Carlo version of the model.

Using [MimiPAGE2009.jl](https://github.com/anthofflab/MimiPAGE2009.jl) as an example, say I want to edit the TCR parameter (i.e. `tcr_transientresponse` [here](https://github.com/anthofflab/MimiPAGE2009.jl/blob/master/src/mcs.jl)). I thought that adding the below line to the `do_monte_carlo_runs()` function might do the trick, but I just get a bunch of error messages:
```
update_param!(mcs, :tcr_transientresponse, Normal(0.0, 0.1)) ## Try crazy low TCR
```
Any suggestions for a better method are much appreciated.

ckingdon95:
Hi Grant, you are correct, the `update_param!` function is meant to be used on a deterministic model, not within the Monte Carlo simulation. If I am interpreting your intention correctly, you would like to sample very low values of TCR throughout the simulation (each trial with a different value from your normal distribution). 

In order to do this, you should modify the definition for the `tcr_transientresponse`  parameter's distribution that is definied within the `@def_sim` macro where all of PAGE's random variables are defined. On line 46 of "src/mcs.jl" in the MimiPAGE2009 package, the random variable is defined as the following:
```
tcr_transientresponse = TriangularDist(1, 2.8, 1.3)
```
Simply change this to your desired distribution
```
tcr_transientresponse = Normal(0.0, 0.1)
```
And then when you run the simulation (using the `do_monte_carlo_runs` function if desired) it will sample from this new distribution instead.

Let me know if you have any other questions about how to use this, I'm happy to help!

grantmcdermott:
Super, thanks Cora. That approach was going to be my fallback, so it's good to have it sanity checked by someone in the know.

grantmcdermott:
PS - It might be helpful to add a note to the docs about the limitations of `!update_param` to the deterministic case. Cheers!

davidanthoff:
[...]

PR #478, opened 2019-05-14
Simulation update funcs

rjplevin:
- Added methods to manipulate a `Simulation` instance
- Eliminated `sim.dist_rvs`
- Changed uses of `sim.dist_rvs` to call `values(sim.rvdict)`

See also my comments in #474

### DIFF
--- a/src/mcs/lhs.jl
+++ b/src/mcs/lhs.jl
@@ -163,18 +163,21 @@ function lhs(rvlist::Vector{RandomVariable}, trials::Int;
     return asDataFrame ? DataFrame(samples, map(rv->rv.name, rvlist)) : samples
 end
 
+function lhs(sim::LatinHypercubeSimulation; 
+             corrmatrix::Union{Matrix{Float64},Nothing}=nothing,
+             asDataFrame::Bool=true)
+    rvlist = collect(values(sim.rvdict))
+    lhs(rvlist, sim.trials, corrmatrix=corrmatrix, asDataFrame=asDataFrame)
+end
+
 function lhs!(sim::LatinHypercubeSimulation; corrmatrix::Union{Matrix{Float64},Nothing}=nothing)
     # TBD: verify that any correlated values are actual distributions, not stored vectors?
 
-    trials = sim.trials
     rvdict = sim.rvdict
-    num_rvs = length(rvdict)
-    rvlist = collect(values(sim.rvdict))
 
-    samples = lhs(rvlist, sim.trials, corrmatrix=corrmatrix, asDataFrame=false)
+    samples = lhs(sim, corrmatrix=corrmatrix, asDataFrame=false)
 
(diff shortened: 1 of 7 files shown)


rjplevin:
These new fns are not used in the code base. I though we might merge them to see if they meet our needs.

ckingdon95:
These functions look great to me! I'm looking at the Travis build though, should we be concerned that something is erroring when trying to do an LHS sample?

test/mcs/test_defmcs.jl
```diff
@@ -11,7 +11,6 @@ using Test
 using Mimi: reset_compdefs, modelinstance, compinstance, 
             get_var_value, OUTER, INNER, ReshapedDistribution
 
-reset_compdefs()
```
ckingdon95: Why is this removed? (I'm not sure I understand why it was there in the first place, but just wondering what's different now)

src/mcs/lhs.jl
```diff
@@ -169,7 +169,7 @@ function lhs!(sim::LatinHypercubeSimulation; corrmatrix::Union{Matrix{Float64},N
     trials = sim.trials
     rvdict = sim.rvdict
     num_rvs = length(rvdict)
-    rvlist = sim.dist_rvs
+    rvlist = collect(values(sim.rvdict))
```
ckingdon95: I just add this extra `collect` because the `lhs` function that gets called below iexpects a vector of RVs, not an iterator. All of the other places where you replace `sim.dist_rvs` with `values(sim.rvdict)` work though because they only get used in for loops

test/mcs/test_defmcs.jl
rjplevin: This follows from the removal of the global list of component defs. Nothing to reset!

davidanthoff:
I do think we should get back into the grove of not merging things that break CI. Can we figure out what is happening there before we merge?

ckingdon95:
@davidanthoff yes there is an actual error, I fixed one of them but there's still another so we definitely should not merge this yet!

ckingdon95:
I just added another spot where we need to use `collect`, but there are now more errors in something to do with GSA. @rjplevin can you take a look? this might involve needing to update something in the GlobalSensitivityAnalysis package, I'm not sure. The error happens during this call to sample on line 32 of  "src/mcs/sobol.jl":
```
    payload = create_GSA_payload(sim)
    samples = GlobalSensitivityAnalysis.sample(payload)
```

rjplevin:
I just posted an update to this branch that passes all Mimi tests.

davidanthoff:
Ready to merge then? @rjplevin do you want to merge, you probably know best whether this is ready?

rjplevin:
I'm testing against all the models... I'll merge it when everything passes!

rjplevin:
All Mimi and model tests are now passing.

PR #482, opened 2019-05-16
added a comment to the to Pkg.add("Mimi")

fhoces:
First time running Julia, and `Pkg.add("Mimi") ` was not working until I did `import Pkg`


davidanthoff:
Thanks for finding this!

@lrennels I think we should instead point folks to the pkg REPL mode and get rid of any code examples that use `Pkg.something`. @lrennels could you track this?

lrennels:
@davidanthoff sure I'll look through the docs now and add a new PR

lrennels:
@davidanthoff @fhoces see #483

lrennels:
This is addressed in #483, so that `import Pkg` will actually not be necessary since use the Pkg mode in the REPL.

PR #501, opened 2019-05-23
Add sim modification functions to userguide

lrennels:


lrennels:
We probably don't, you're right and I was thinking about that, but wanted to link to the docstrings in which case I needed them in the `Reference.md` file.  Perhaps I won't make the `@ref` links and users can go look them up if they're curious.

I realized in the meantime that our `Reference.md` file isn't building properly ...

lrennels:
Ah I think I found the problem there looking at the `v0.9.0` docs build ... it can't the `new_comp` doctstring

PR #513, opened 2019-05-28
Fix bug in dotview() for timesteps

rjplevin:
Fix bug in dotview() for timesteps and added preliminary example sent by James as test/test_broadcast.jl (not fully wired into testing system yet...)


rjplevin:
Ok -- it doesn't seem to break anything, and James hasn't declared it broken, so I'll merge.

PR #518, opened 2019-06-04
Explore Simulation

lrennels:
Issue #477


lrennels:
Note that need to update for the new sim API and pre-calculate statistics for trumpet plots

lrennels:
Some TODOs:

- pre-filter data for min/max/mean before passing to `VegaLite` ... and come up with similar method for histograms
- add methods for specifying an `output_dir`

lrennels:
@davidanthoff I'm having trouble debugging why calls to `plot` are working for my trumpet plots, but calls to `explore` are not displaying the plot properly.  This isn't a problem with `explore(m)` but maybe it has to do with the syntax I'm using to create these plots.

Both use the exact same function to get the `spec` (`Mimi._spec_for_sim_item`), which holds the full spec. `plot` then calls `return VegaLite.VLSpec{:plot}(spec[:VLspec])` where `:VLspec`has the actual specification. In contrast, `explore` passes the full JSONified `JSON.json(spec)` to a Javasript function `display`, which has the signature:

```
function display(spec) {
    vegaEmbed("#vis", spec["VLspec"], {actions: false});
}
```

For some reason, however, `plot` displays things perfectly, while `explore` has trouble rendering the full plot.  Are there differences that you know of between VegaLite. `vegaEmbed` and VegaLite.jl's `VegaLite.VLSpec{:plot}` that could cause a difference here?  I'm a bit stuck ...

lrennels:
Screen Shot 2019-06-26 at 5 50 43 PM
Screen Shot 2019-06-26 at 5 38 14 PM

lrennels:
But the `explore` function  gives me
Screen Shot 2019-06-26 at 5 51 12 PM
and
Screen Shot 2019-06-26 at 5 51 20 PM

lrennels:
Also to handle ...

- [ ] regional trumpet plots overflow the `div`
- [ ] regional trumpet plot interactivity overlaps between graphs ... 
- [ ] histograms and regional histograms

lrennels:
@rjplevin if you get the chance take a look at this PR and let me know if you have any thoughts.  It's a first go and an `explore(sim_inst::SimulationInstance)` function and the partner `plot` functions.  I'll write some documentation and graphics soon, but in essence there are:

lrennels:
- [ ] documentation
- [ ] `output_dir` argument methods
- [ ] the `transform` to filter interactive plots in `explore(model)` doesn't work now, only the scale way we do it in `explore(sim)` but that's not ideal for the former ... both now use `domain` but `transform` would be ideal

PR #520, opened 2019-06-08
Add support for creating an empty Simulation

rjplevin:
Made it possible to call `sim = Simulation()` and use the new API to populate `sim`.

### DIFF
--- a/src/mcs/defmcs.jl
+++ b/src/mcs/defmcs.jl
@@ -165,7 +165,7 @@ macro defsim(expr)
         # TBD: need to generalize this to support other methods
         return :(Simulation{$simdatatype}(
                     [$(_rvs...)],
-                    [$(_transforms...)],  
+                    TransformSpec[$(_transforms...)],  
                     Tuple{Symbol, Symbol}[$(_saves...)],
                     $data))
     end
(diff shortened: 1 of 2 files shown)


lrennels:
What’s the use case here? Doesn’t the new API work about the same but by deep copying the existing sim and then adding a few characteristics like models and N?

rjplevin:
This was in response to James' issue. It occurred to me that he could allocated an empty sim and use the functional API to add to it in each component's file. We might streamline the API as we discussed. 

Alternatively, we could modify `@defsim` so it works on a pre-allocated `Simulation`. This would provide the "incremental defsim" we talked about today. It would basically do what I suggested above, but it would be consistent with the existing approach. In fact, it would be a handy way to modify an existing `Simulation`, say, returned from `Mimi.get_sim()`.

lrennels:
Ah ok so this is in line with our conversation at the end today, not just the API we had on the board.

davidanthoff:
We could also generally think about support for merging simulations? Either via `merge(s1, s2)`, or maybe even something like `s1 + s2`? Although I kind of like the first option better, because it is more explicit that this operation is not commutative.

rjplevin:
Yes, `merge(s1, s2)` seems like the ticket. We could also support `merge!(s1, s2)`, as for Dicts.

lrennels:
Looks good to me!

Issue #521, opened 2019-06-09
Simulation API (Consolidated)


lrennels:
PR #522 

- [x] Simplify`run_sim` function to allow for the following (and adjust all helper functions etc., accordingly). Importantly note that we have moved `models` and `results` out of the `Simulation` instance and into the `SimulationResults` instance returned by `run`.  The `SimulationResults` instance holds (1) a (mutated) copy of the passed `Simulation` (2) a list of `models` or `model` (3) a Dictionary of `results`.

```julia
sim = get_a_sim_def()
run(sim, m, ntrials=1000) 
```

- [x] Add access to results using `sim_inst[:foo, :bar; model = 1] => DF`

- [x] Add results streaming to disk

lrennels:
This update so far includes removing the `trials` and `models_to_run` keyword arguments from the `run` function, since they do not make as much sense with the new API.  We should consider how/if to reimplement these.  We also should consider if callback functions should take `Simulation` or `SimulationResults`.  If they take `Simulation`, then we need to move `models` inside the `Simulation` held by `SimulationResults` so that the callback functions can access the model list.

lrennels:
Noting another question: should `SimulationResults{T}` be parameterized by `<: AbstractSimulationData` or `<:Simulation{T1} where T1 <: AbstractSimulationData`?

lrennels:
@rjplevin did you want to perhaps take a look at the simulation modification functions like `addRV!` etc. to see if you wanted to simplify that API in the short-term that we can push out with this PR and documentation?

PR #522, opened 2019-06-10
Improved Simulation API

lrennels:
for Issue #521 

PSA on the Forum: https://forum.mimiframework.org/t/psa-simulation-api/69

@grantmcdermott I tagged the MCS users I know of on the forum, but I don't think you've used it yet so I couldn't tag you ... take a look!


lrennels:
Some Notes:

- This update removes the `trials` and `models_to_run` keyword args from the `run` function, since they do not make as much sense with the new API.  If they become desirable we can discuss reimplementing.  
- This update changes callback functions to take  `SimulationResults`, not a `Simulation` in their arguments.  This way they can access the `models` vector held by `SimulationResults`.  
- `SimulationResults{T}` is currently parameterized by `<: AbstractSimulationData`, but could be parameterized by  `<:Simulation{T1} where T1 <: AbstractSimulationData`
- `run` now has a keyword argument `results_in_memory::Bool = true` so that unless it is set to `false` results will be stored in the `results` Dictionary, _and_ in the `results_output_dir` if and only if it is specified.  
- **BUG: We need to add functionality for the in-memory `results` dict to handle multiple scenarios though ... it currently overwrites for each scenario. We could handle the logic in a similar way to the folders structure of the results ... @rjplevin thoughts or help on this one?**
- To access results with `getindex` we will also need to indicate which `model`, and potentially the scenario ... decide how to do this after we pick a `results` structure (bullet above)

docs/src/internals/montecarlo.md
```diff
@@ -2,15 +2,11 @@
 
 ## Overview
 
-Sensitivity Analysis support consists of four primary user-facing elements:
+Sensitivity Analysis support consists of two primary user-facing elements:
```
rjplevin: > `consists of two primary user-facing elements:` 

Change to `consists of the following user-facing elements:` and avoid having to change this again!

test/mcs/test_defmcs.jl
```diff
@@ -92,7 +77,7 @@ end
 #
 global loop_counter = 0
 
-function outer_loop_func(sim::Simulation, tup)
+function outer_loop_func(sim_results::SimulationResults, tup)
```
rjplevin: I can see use cases for loop funcs that use the trial data held in the `Simulation`. We might want to change the API to these to be `outer_loop_func(sim::Simulation, results::SimulationResults, tup)` now that we've split the original struct.

test/mcs/test_defmcs.jl
```diff
@@ -135,20 +119,20 @@ run_sim(sim;
...
 
 
-function other_loop_func(sim::Simulation, tup)
+function other_loop_func(sim_results::SimulationResults, tup)
     global loop_counter
     loop_counter += 10
 end
 
-function pre_trial(sim::Simulation, trialnum::Int, ntimesteps::Int, tup::Tuple)
+function pre_trial(sim_results::SimulationResults, trialnum::Int, ntimesteps::Int, tup::Tuple)
```
rjplevin: Same as comment re: loop funcs. Pass both `Simulation` and `SimulationResults`.

rjplevin:
I now think `SimulationResults` is not the best description of what we have here. Two things bothered me as I read through Lisa's (otherwise excellent!) changes:

1. `SimulationResults` has a field `results`. The other elements in the struct (i.e., `sim`, `models`) are not "results", so there's a concept/name mismatch here.

2. This is exemplified by calling the `pre_trial_func` with `SimulationResults`. That seems wrong, doesn't it? Sure, the struct provides access to a copy of the `Simulation`, but see #1. ;~)

Naming is [hard](https://martinfowler.com/bliki/TwoHardThings.html), of course. Should we rename the old `Simulation` => `SimulationDef`, and the current `SimulationResults` => `SimulationInstance` in keeping with the rest of Mimi?

lrennels:
Ah interesting I see the issue there, I would be ok with `SimulationDef` and `SimulationInstance`, that seems consistent. I also really didn't like the callback functions being called on `SimulationResults` but also didn't want to stick `models` into the `Simulation`, so perhaps this helps to clear up that problem.

lrennels:
From 6/11

- To save results in files, create a tree structure of scenario --> model --> variable like we do now.  Right now we have an error here where trials are overwritten, which is a problem in `_store_trial_results, so we need to fix that and then implement streaming with `CSVFiles.save_streaming(filename, file)` and `CSVFiles.save_streaming(stream, file)` ... see tests for information on how to do this.
- To access results internally we can write `getindex` and `getdataframe` to flatten out the above.
- To save results internally, use a multi-dimensional array with an extra dimension for model or scenario
-

src/mcs/montecarlo.jl
```diff
@@ -14,56 +16,73 @@ function print_nonempty(name, vector)
...
             results_df = results[datum_key]
         else
-            results_df = DataFrame([typeof(value), Int], [datum_name, :trialnum], 0)
+            if has_scen
+                results_df = DataFrame([typeof(value), Int, String], [datum_name, :trialnum, :scen], 0)
+            else
+                results_df = DataFrame([typeof(value), Int], [datum_name, :trialnum], 0)
+            end
             results[datum_key] = results_df
         end
```
rjplevin: This is probably a matter of taste, but in my experience redundancy always (eventually) creates more work because you have to change multiple things rather than one. I'd write this as follows, which slightly longer but without redundancy:
```
types = [typeof(value), Int]
names = [datum_name, :trialnum]
if has_scen
   push!(types, String)
   push!(names, :scen)
end
results_df = DataFrame(types, names, 0)
results[datum_key] = results_df
```

src/mcs/montecarlo.jl
```diff
@@ -75,106 +94,118 @@ function _store_param_results(m::Model, datum_key::Tuple{Symbol, Symbol}, trialn
...
-            filename = joinpath(sub_dir, "$datum_name.csv")
-            save(filename, results[datum_key])
-        end
+function _save_trial_results(trial_df::DataFrame, datum_name::String, output_dir::AbstractString, streams::Dict{String, FileIO.Stream}) where T <: AbstractSimulationData
+    filename = joinpath(output_dir, "$datum_name.csv")
+    if haskey(streams, filename)
+        CSVFiles.savestreaming(streams[filename], trial_df)
+    else
+        streams[filename] = CSVFiles.savestreaming(filename, trial_df)
     end
```
rjplevin: This should work without the `CSVFiles.` before `savestreaming`. Without specifying `CSVFiles`, we can optionally save to any recognized file format that supports streaming.

src/mcs/montecarlo.jl
```diff
@@ -363,57 +400,58 @@ If provided, `scenario_args` must be a `Vector{Pair}`, where each `Pair` is a sy
...
+    if (!results_in_memory) && (results_output_dir===nothing)
+        error("The results_in_memory keyword arg is set to ($results_in_memory) and 
+        results_output_dir keyword arg is set to ($results_output_dir), thus 
+        results will not be saved eitehr in memory or in a file.")
+    end
+
+    # Initiate the SimulationInstance and set the models and trials for the copied 
+    # sim held within sim_inst
+    sim_inst = SimulationInstance{typeof(sim_def.data)}(sim_def)
+    set_models!(sim_inst, models)
```
rjplevin: If you define `SimulationInstance` in terms of a type `T`, you can avoid this explicit call to `typeof(sim_def.data)`. For example:
```
julia> struct SimDef{T}
         data::T
         end

julia> struct SimInst{T}
         sim_def::SimDef{T}
         end

julia> sd = SimDef((10, 20))
SimDef{Tuple{Int64,Int64}}((10, 20))

julia> SimInst(sd)
SimInst{Tuple{Int64,Int64}}(SimDef{Tuple{Int64,Int64}}((10, 20)))
```

lrennels: So given I already have this:

rjplevin: Yes, and I think you can drop the where clause from
```
sim_def::SimulationDef{T} where T <: AbstractSimulationData
```
since the constructor enforces this.

lrennels:
@davidanthoff we will be ready to merge this, and then merge the changes to the models and rebuild Mimi so all dependency tests pass, after CSVFiles `savestreaming` PR is merged so order of operations:

1) merge @rjplevin 's `savestreaming` branch in `CSVFiles`
2) merge this PR (I can do that or you can)
3) merge `simapi` PRs for models
4) tag new `Mimi` release (I can do that or you can), which will rebuild and tag the new docs etc.

rjplevin:
Nice work! I appreciate your attention to detail and process.

src/mcs/montecarlo.jl
lrennels: Ah ok, I tried this and got some complaints about my `SimulationInstance` call to `new` but I'll play with it a bit more, the elusive `{T}` syntax ...

lrennels:
aka minor but useful OCD

src/mcs/montecarlo.jl
rjplevin: You might need `new{T}(args...)`

lrennels:
@rjplevin I just moved `trials`, `current_trial`, and `current_data` to `SimulationInstance` from `SimulationDef`. Can you take a look at that particular commit, and specifically the **changes to functions and iteration support** to make sure it seems like the cleanest/most straightforward way to represent things?  We could always add methods that take a `sim_inst::SimulationInstance` and call the function on the `sim_inst.sim_def` in some cases if it's cleaner.

rjplevin:
Does this work? 
```julia
IteratorInterfaceExtensions.getiterator(sim_inst::SimulationInstance) =
   SimIterator{sim_inst.sim_def.nt_type}(sim_inst)
```

I would expect the `T` is required, i.e.,

```julia
IteratorInterfaceExtensions.getiterator(sim_inst::SimulationInstance{T}) where T = 
    SimIterator{sim_inst.sim_def.nt_type, T}(sim_inst)
```

Or, we could define an "outside" constructor:

```julia
SimIterator(sim_inst::SimulationInstance{T}) where T = 
   SimIterator{sim_inst.sim_def.nt_type, T}(sim_inst)
```
and then simply:
```julia
IteratorInterfaceExtensions.getiterator(sim_inst::SimulationInstance) = SimIterator(sim_inst)
```

I don't remember why the iterator needs to be parameterized on`NT`, since this is always available as `sim_inst.sim_def.nt_type`. Could be an artifact. I'd say leave it for now and we can see about removing it later.

lrennels:
@rjplevin it does seem to work without the `T`, while
```
IteratorInterfaceExtensions.getiterator(sim_inst::SimulationInstance{T}) where T = 
    SimIterator{sim_inst.sim_def.nt_type, T}(sim_inst)
```
errors with 
```
LoadError: MethodError: no method matching Mimi.SimIterator{NamedTuple{(:name1, :name2, :name3, :share!2, :sigma!3, :depk!4, :depk!5, :depk!6),NTuple{8,Float64}},LHSData}(::SimulationInstance{LHSData})
  Stacktrace:
```

rjplevin:
I guess the underlying MCSData type associated with `sim_inst` is carried through to the SimIterator. I expected the explicit version of this to work in any case. Is this pushed? Hard to say more without seeing the code.

lrennels:
The explicit version being the first option with`T`?

lrennels:
Just pushed that one, the failing tests start in `test_defmcs.jl`

lrennels:
Ah ok I think I figured it out, our `SimIterator` was written like this:

rjplevin:
Cool.

PR #523, opened 2019-06-10
Fix bug in inner scenario results saving

lrennels:
@rjplevin I think this was a bug in saving results, only the results for the last loop were being saved in the last of the tuples ... I'm fixing it it the bigger PR for the Sim API but thought we should just push out the bug fix now too

### DIFF
--- a/src/mcs/montecarlo.jl
+++ b/src/mcs/montecarlo.jl
@@ -493,12 +493,12 @@ function run_sim(sim::Simulation{T};
                 _restore_sim_params!(sim, original_values)
 
                 counter += 1
-                ProgressMeter.update!(p, counter)                
-            end
-
-            if has_inner_scenario && has_output_dir
-                save_trial_results(sim, output_dir)
-                _reset_results!(sim)
+                ProgressMeter.update!(p, counter)  
+                
+                if has_inner_scenario && has_output_dir
+                    save_trial_results(sim, output_dir)
+                    _reset_results!(sim)
+                end              
             end
         end
 


rjplevin:
Hi, Lisa. I'm confused by your saying "you think this was a bug" -- why the uncertainty?

lrennels:
Only because you wrote the code and I wanted to double check that this is the  right fix etc. and you didn’t intend anything else.  95% sure :)

rjplevin:
No worries! For the record, there's no need to be polite about fixing bugs! (At least not mine.) You can't write software without writing bugs. Fixing them is always a good thing!

The other thing I was unsure about was whether the bug was in storing the results or writing the results to disk. (Or both...)

lrennels:
Just with writing results to disk! The writing to disk happened one loop too far out

rjplevin:
It looks to me like the bug is not in the placement of `save_trial_results` but in `_store_trial_results`, which doesn't "know" about scenarios, so it overwrites the prior scenario's results each time.

The PR saves trial results not at the end of the trial (after all the inner scenarios for that trial complete), but after each inner scenario, then resets the results. Isn't this wiping out previously saved results?

I think we need to store results differentiated by scenario, which could just be the number of the scenario, much as we do with results from multiple models.

It also looks like we're writing the full results file after each trial when we should be appending after each trial. I don't know if `save()` handles this. The first write would need to be a special case that writes out column headers.

Let's discuss tomorrow. I'd be happy to help with this, or take it back (since these are my bugs!) as you prefer.

lrennels:
Yes, this was sort of a temporary fix for a larger problem.  We need to solve the fact that `_store_trial_results` does not handle scenarios, so for now I figured moving it would at least solve mis-saved files ... but you are correct it overwrites the stored results which is the main bug I mention in the Sim API PR.

rjplevin:
Do you want to keep going with this or kick it back to (at!) me?

lrennels:
Secondly, we have the problem of `_store_trial_results` not saving proper structures in the `results` Dictionary and overwriting scenarios.  I point that out in the Sim API.

rjplevin:
I see now that you pointed it out earlier!

lrennels:
I can kick this part back to you.  I'd say if you're going to fix both problems I list able it might be best to just do it in the Simulation API PR, as long as we keep that moving and merge it sooner rather than later.

lrennels:
But fixing it in either place would really be fine... whatever organization makes sense to you?  I was going to do the naming changes to `Instance` and `Definition` now

rjplevin:
Why don't you keep going for now. I'll think about it but not push any changes yet.

lrennels:
Alright sounds good. Tomorrow we can discuss how to properly save results in-memory when scenarios are involved, and also how we want to `getindex`... because we need a few extra parameters to index into results when we have multiple models and scenarios.

lrennels:
I'll close this for now and keep it all in the Sim API branch

PR #532, opened 2019-07-28
comp_id.module

ckingdon95:
@davidanthoff this is for what we were talking about on Friday, to have the `comp_id` in a component definition hold the full module instead of just a symbol of the bottom most module. This currently works now (on the MimiIWG example that wasn't working before) and passes all of the Mimi tests (although I had to change three tests that were specifically testing this functionality).

Would it be better though to have the `comp_id` still store a symbol, but just make sure the symbol is the full module path (i.e. make it hold `Symbol("Main.foo.bar")` instead of what it would previously store, which was just `:bar`)? I'm just not sure what the implications are of passing around the actual module object and if there are any disadvantages of this we should be worried about.


ckingdon95:
I put an example of the problem in an issue here: #533

davidanthoff:
I think @rjplevin needs to make the call here. To me it seems fine to store the module itself instead of the name. With the caveat that we should think through what that means if someone redefines a module... Probably it would mean that one would then have a reference to an outdated module around...

ckingdon95:
Rich started implementing this differently in his `class-based` branch as well, to have it store the whole module path as symbols instead. We should remember to talk about this together on Monday!

ckingdon95:
@davidanthoff I believe we decided to merge this as is for now. Do you want to take a look before I merge it? The travis tests aren't passing because the models are on the old versions

davidanthoff:
Looks good!

We should really try to move to a situation where we don't end up with a broken CI build on `master`, which seems to be what is going on here, right?

ckingdon95:
Yes, I'm not sure how to prevent that when we have updates with breaking changes. I think Lisa has a better grasp on what the ideal workflow could be. One thing on this PR before I merge it though: should I rename the field in the `comp_id` to be called `module` instead? Or is it okay to leave it as `module_name`?

davidanthoff:
> should I rename the field in the comp_id to be called module instead?

PR #546, opened 2019-09-12
Update manifest

lrennels:


lrennels:
@davidanthoff can you take a look at my (numerous) attempts to get the dependencies tests passing ... I have it working for some tests with `#master` but that might not be what we want and I'm still failing a few ...

PR #550, opened 2019-09-19
WIP Support for MarginalModels in running simulations

ckingdon95:
These changes allow `MarginalModels` in a simulation instance's list of models. Previously, a `MarginalModel` `mm` would be "flattened" into two models `mm.base` and `mm.marginal` in the list of models. Now, if a user provides a MarginalModel to the `run` function, then accessing the list of models from a simulation instance will yield the exact same list of models and/or MarginalModels. 

This functionality is demonstrated in a new test file: "test/mcs/test_marginalmodels.jl".

### DIFF
--- a/src/core/build.jl
+++ b/src/core/build.jl
@@ -190,3 +190,8 @@ function Base.run(mm::MarginalModel; ntimesteps::Int=typemax(Int))
     run(mm.base, ntimesteps=ntimesteps)
     run(mm.marginal, ntimesteps=ntimesteps)
 end
+
+function build(mm::MarginalModel)
+    build(mm.base)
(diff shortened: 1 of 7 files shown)


lrennels:
@davidanthoff something is wrong with the Travis check that might have to do with `Excel` deployment on 1.2?

davidanthoff:
Yes, I just restarted the build, I think it might have been a temporary download issue.

davidanthoff:
Hm, this is annoying: even a failing entry in the build matrix that is allowed to fail seems to prevent merging... @ckingdon95 should I merge this? Tests pass.

Issue #549, opened 2019-09-15
Allow indexing into a Variable with a year


lrennels:
Conversation started here: https://github.com/anthofflab/MimiMAGICC.jl/issues/2

davidanthoff:
Worked on in https://github.com/mimiframework/Mimi.jl/pull/551.

Issue #337, opened 2018-10-19
better timestep comparison handling run_timestep


lrennels:
Think of a clean way to handle the types of comparisons in `run_timestep`

- the timestep is ==, >, < a year
- the time is ==, >, > an int
- perhaps an easier way to access `t.t` (@FrankErrickson)

ckingdon95:
So for this one, we still want to add `<, >, =` comparisons for the new `TimestepIndex` and `TimestepValue` types to compare with `AbstractTimesteps` right?

lrennels:
Yep I think that's right!  It's on me, I just haven't gotten to it.

PR #551, opened 2019-09-19
WIP: Allow Indexing into a Variable/Parameter with a Year

lrennels:
UPDATE: CLOSING THIS PR TO START FRESH

#549, #568, #337 

This PR will handle a few different Issues related to indexing into a Variable or Parameter.

- [ ] create the two new types
- [ ] add `Base.:+` and `Base.:-` methods for the two types
- [ ] add `getindex` methods to `TimestepArray`, `TimestepVector`, and `TimestepMatrix`
- [ ] deprecate int indexing (#568)
- [ ] take a look at #337 
- [ ] `time_index`
- [ ] tests and documentation

### DIFF
--- a/src/core/time.jl
+++ b/src/core/time.jl
@@ -125,6 +125,14 @@ function Base.:+(ts::VariableTimestep{TIMES}, val::Int) where {TIMES}
 	new_ts = VariableTimestep{TIMES}(ts.t + val)
 end
 
+function Base.:-(ts::Timestep, val::Int) 
+	return Timestep(ts.value, ts.offset_in_timesteps - val)
+end
+
+function Base.:+(ts::Timestep, val::Int) 
+	return Timestep(ts.value, ts.offset_in_timesteps + val)
+end
+
 #
 #  2. CLOCK
 #
@@ -240,15 +248,19 @@ function Base.getindex(v::TimestepVector{VariableTimestep{D_TIMES}, T}, ts::Vari
 end
 
 function Base.getindex(v::TimestepVector{FixedTimestep{FIRST, STEP, LAST}, T}, ts::Timestep) where {T, FIRST, STEP, LAST} 
-
-	t = findfirst(isequal.(ts.value + ts.offset_in_timesteps, [first:step:last...]))
-	t === nothing ? error("cannot get data for year $(ts.value), time is not in the list of time values "): data = v.data[t]
+	t = findfirst(isequal.(ts.value + ts.offset_in_timesteps, [FIRST:STEP:LAST...]))
+	if t === nothing
+		error("cannot get data for year $(ts.value + ts.offset_in_timesteps), time is not in the list of time values ")
+	end
+	data = v.data[t]
 	_missing_data_check(data)
 end
 
 function Base.getindex(v::TimestepVector{VariableTimestep{TIMES}, T}, ts::Timestep) where {T, TIMES}
 	t = findfirst(isequal.(ts.value + ts.offset_in_timesteps, TIMES))
-	t === nothing ? error("cannot get data for year $(ts.value), time is not in the list of time values "): data = v.data[t]
+	if t === nothing
+		error("cannot get data for year $(ts.value + ts.offset_in_timesteps), time is not in the list of time values ")
+	end
 	data = v.data[t]
 	_missing_data_check(data)
 end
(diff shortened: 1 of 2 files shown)


lrennels:
@davidanthoff @rjplevin here's a first pass at these, I will go back over it and see ifi I can make things clearer or faster, as well as add tests and docs, but for now wanted to get the general ideas down and see what you guys think?

davidanthoff:
Just to record what we decided in our meeting today: we'll have two new types, one called `TimestepValue` and one called `TimestepIndex`. And we'll wait with this PR until the composite components have been merged into `master`.

ckingdon95:
I was still feeling unsure about these type names, `TimestepValue` and `TimestepIndex` since the thing that gets constructed is actually a `Timestep`, not a value or an index. What if instead we offer two keyword options: `Timestep(value = 1952)` or `Timestep(index = 2)`? Not super important, but was wondering what you guys thing of that instead. (I'm also not even sure if you can have keyword arguments like that in a constructor) @davidanthoff @lrennels @rjplevin

rjplevin:
@ckingdon95, I see it slightly differently. The problem is that we have two conceptually different arguments, both integers, that we want to distinguish. If we define: 
```
struct TimestepIndex
  value::Int
end

struct TimestepValue
  value::Int
end
```
we've effectively created two new type-system-distinguishable integers. So I don't see `TimestepIndex(2)` as producing a `Timestep`, but producing an integer value of 2 "tagged" to be interpreted as an index rather than as a year.

ckingdon95:
Ah I see. If we are creating two different types then that makes sense. I thought that both of these would create the `Timestep` type that Lisa added and that both would hold the `value` and `offset_in_timesteps` fields. If we're creating the two different types you've shown here though, is it obvious how we resolve the arithmetic operations we discussed?

davidanthoff:
We should probably make it
```julia
struct TimestepValue{T}
  value::T
end
```
so that one isn't limited to `Int` for the value domain here, but could instead also use a `Date` or something, if one wanted to.

rjplevin:
Agreed about type parameterization.

As we've discussed, a time value, e.g., 2005, cannot be used without reference to the actual time dimension definition. So I don't see how `TimestepValue(2005)` would be converted to a timestep. It could, however, be _interpreted_ w.r.t. a time dimension, as could `TimestepIndex(2)`.

If we define these as you discussed on the call, with an offset field, e.g.,
```julia
struct TimestepValue{T1, T2}
  value::T1
  offset::T2
end

struct TimestepIndex{T1, T2}
  value::T1
  offset::T2
end
```
then the add/subtract functions just operate on the offset. The whole thing is ultimately interpreted w.r.t. a given time dimension.

davidanthoff:
Yes, exactly, `TimestepValue` can only be resolved to an index in combination with a time array, which has the necessary offset info.

And yes, I think the fields in the `TimestepIndex` must be `Int`. Same for the `offset` in `TimestepValue`.

ckingdon95:
Does `TimestepIndex` even need the `offset` field then? Can't we just do arithmetic by incrementing it's `value`?

rjplevin:
That sounds right. I suppose we could also say that if it's an `Int`, it's an index, and treat the `TimestepValue(x)` differently. I do prefer the symmetry and lack of ambiguity of the two types, though.

ckingdon95:
Maybe it should be:
```julia
struct TimestepIndex
    index::Int
end
```
Even less symmetrical, but maybe more clear?

davidanthoff:
Yes, I think that is how it should look.

lrennels:
This looks good to me, although @davidanthoff when you say that we can type parameterize `TimestepValue` with something like `Date`, that would only work if we add functionality so that the `TimestepArray` holds `Date` types instead of `Int` types right?  Because right now this function is key to finding the index, but obviously uses a lookup in the array:

```
# Helper functions for Timestep type
function _get_time_value_position_(times::Array, ts::Timestep)
	t = findfirst(isequal.(ts.value, times))
	if t === nothing
		error("cannot use Timestep with value $(ts.value), value is not in the TimestepArray")
	end
	
	t_offset = t + ts.offset_in_timesteps
	if t_offset > length(times) 
		error("cannot get offset of $(ts.offset_in_timesteps) from $(ts.value), offset is after the end of the TimestepArray")
	end
	return t_offset
end
```

ckingdon95:
I think it has to do with what type is used when the user calls `set_dim!(m, :time, ...)`. So far, all our models have `Vector{Int}` as the time dimension, but it could be Dates. I think there are places in the code that would have to be updated to support this, but I think this `_get_time_value_position_` function should still work, since I'm assuming `isequal` work for Dates, and the `times` parameter would be a vector of Dates if that was the specified type.

lrennels:
Agreed!

ckingdon95:
@lrennels should we also add a `getindex(t)` function (similar to the `gettime` function) that returns the index value? This wouldn't work on `TimestepValue` objects, but should work on `TimestepIndex`, `FixedTimestep` and `VariableTimestep`. I think the use of this would just be in the run_timestep function if someone wanted to know if they were on a certain timestep other than the first or last, e.g. the second timestep, like `getindex(t) == 2`. I vaguely remember this being discussed somewhere else, but I can't remember where, so sorry if this was already resolved!

rjplevin:
We should call it something else since `Base.getindex` is the array access method. Maybe `time_index`?

davidanthoff:
I think I would actually prefer that such a test is written as `t == TimestepIndex(2)`, rather than us providing a way to extract the index value. On some level, we should really discourage the use of these index values as much as possible...

PR #556, opened 2019-09-30
Class based

rjplevin:
The long awaited composite mega-super-mondo-merge.


davidanthoff:
I think the next step here is to modify https://github.com/mimiframework/Mimi.jl/blob/master/test/dependencies/run_dependency_tests.jl such that it uses a branch of FUND that works with this version of Mimi.jl. I think simply replacing the second element in the tuple that is currently a version string should do the trick.

Or, alternatively we merge the FUND branch, tag a new release, and then also update that file here to pick up the newer version.

PR #557, opened 2019-09-30
Use Doctests

lrennels:
This PR addresses issue #548.

Adding Documenter doctests to the Mimi documentation so that the code blocks are run during testing.  Doctests will be included as much as possible, including in the tutorials, but for now are not used comprehensively in files such as the `userguide.md`, since that would require substantial configuring of the file (it has most syntax instructions that don't have enough setup code to fully run). 

- [x] primary files
- [x] tutorials 
   - [x] tutorial 1 - `nsteps` error fixed in new FUND version
   - [x] tutorial 2
   - [x] tutorial 3
   - [x] tutorial 4
- [x] internals - no inclusion of doctests in these files for now


docs/src/faq.md
```diff
@@ -1,5 +1,18 @@
 # Frequently asked questions
 
+```@meta
+DocTestSeup =
+    using Pkg
+    Pkg.add("Mimi")
+    Pkg.add("DelimitedFiles")
+    Pkg.add("Distributions")
```
davidanthoff: We shouldn't install the packages in here, instead they should be added to [this file](https://github.com/mimiframework/Mimi.jl/blob/master/docs/Project.toml), then they will be installed automatically for the doc build.

lrennels: Ah ok good I was going to ask about this.  I need to put a little thought into how to incorporate these tests, with one-off examples they make sense but in piecewise tutorials with text in between code blocks it’s a little less obvious how to do this efficiently.

lrennels: @davidanthoff should the `[compat]` line of the Project.toml file have a line for everything in `[deps]`?

lrennels: as long as it's not in base

docs/src/faq.md
```diff
@@ -28,7 +44,7 @@ Here, `../data/cubicparams.csv` is a parameter definition file that looks someth
 ```
 # Example estimated polynomial parameter
 # First line: linear, quadratic, cubic
-# Lines 2-4: covariance matrix
+# Lines 2-5: covariance matrix
```
ckingdon95: I think 2-4 is correct?

ckingdon95: But also, we don't support this do we? passing a distribution to `set_param!`? Is this a very old example?

lrennels: Yes, this is the example that I was going to show @davidanthoff to see if I should take it out completely.  It isn't allowed right now, and I'll be slowly going through all docs to add tests and that'll help us make sure everything is up to date.

docs/src/faq.md
```diff
@@ -11,21 +17,44 @@ covariance between estimated coefficient parameters.  For example, an estimated
 polynomial can be represented as a multivariate Normal distribution,
 with a variance-covariance matrix.  To use this, define the parameter
 in the component with a vector type, like here:
```
lrennels: @davidanthoff should I remove this FAQ question all together since we no longer support this?

lrennels:
@davidanthoff for the tutorials, using doctests would require pulling down models from our registry. Is there a particular way you would suggest doing that? I don't think I can do it in the Project.toml file, but thought maybe I could do something similar to what you do for testing dependencies or something?

docs/src/faq.md
davidanthoff: Yes, I think the docs should always match what we ship!

davidanthoff:
Hm, good question... I guess we could add the registry [here](https://github.com/mimiframework/Mimi.jl/blob/master/.travis.yml#L37), and then the models can just be listed in the `Project.toml` in the `docs` folder?

lrennels:
Ah that's a good idea, do you know the command when you aren't in the `>pkg` REPL?  We'd do `using Pkg` and then `Pkg.add( ? )` ... is there a `registry` keyword or something?

docs/src/tutorials/tutorial_1.md
```diff
@@ -30,34 +36,43 @@ In order to run FUND, you will need to have the packages `Distributions` and `St
...
+In this case, the function `get_model` has the signature
+```julia
+MimiFUND.get_model(; nsteps = default_nsteps, datadir = default_datadir, params = default_params)
 ```
 Thus there are no required arguments, although the user can input `nsteps` to define the number of timesteps (years in this case) the model runs for, `datadir` to define the location of the input data, and `params`, a dictionary definining the parameters of the model.  For example, if you wish to see only the first 100 timesteps,you may use:
-```
+```julia
 using MimiFUND
-m = getfund(nsteps = 100)
+m = MimiFUND.get_model(nsteps = 100)
```
lrennels: @ckingdon95 I'm seeing an error after this call now:

```
ERROR: Mismatched data size for a parameter connection: dimension :time in <ComponentId MimiFUND.scenariouncertainty> has 101 elements; external parameter :scenpgrowth has 1051 elements.
```

any ideas?

docs/src/tutorials/tutorial_4.md
```diff
@@ -246,45 +262,59 @@ Case: We want to do an SCC calculation across a base and marginal model of `Mimi
...
+base = MimiDICE2010.get_model()
 
-#Build the marginal model, which here involves a dummy function `construct_marginal_dice()` that you will need to write
-marginal = construct_marginal_dice(year) 
+#Build the marginal model
+mm = MimiDICE2010.get_marginal_model(m, year = 2015)
 
 # Run
-si = run(sd, [base, marginal], N; trials_output_filename = "ecs_sample.csv", post_trial_func = my_scc_calculation)
+si = run(sd, [mm.base, mm.marginal], N; trials_output_filename = "ecs_sample.csv", post_trial_func = my_scc_calculation)
```
lrennels: @ckingdon95 I know this isn't quite right anymore, since this line `si = ...` fails. Can you take a look and see if there's an easy fix or if I should delete this example?

ckingdon95: Is it okay if I push to this branch? I think there are a couple things wrong. It looks like the line 311 should construct the marginal model from `base` not `m`, and I think there's also a problem in the `my_scc_calculation` function definition (there is no `:DAMAGES` variable in `:neteconomy`, we should be using the consumption levels for the calculation)

docs/src/tutorials/tutorial_1.md
ckingdon95: This PR address this and needs discussion:
https://github.com/fund-model/MimiFUND.jl/pull/44

docs/src/tutorials/tutorial_4.md
lrennels: @ckingdon95 yes go for it that would be great! I’ll handle the doctests parts after the code looks right.

docs/src/tutorials/tutorial_1.md
lrennels: Ah ok thanks!

lrennels: @ckingdon95 I’m realizing this actually might work in real cases since it’s case 2 (the user set a shorter parameters dictionary) ... I have that in this code but don’t run that reset through the doctests code blocks because I don’t have an actual shorter set of parameters on hand.  If I’d done that I bet it would’ve worked.

ckingdon95: @lrennels should we change this `nsteps` example since the fix for this isn't merged into FUND yet? Or should we wait? Ideally these blocks would also be doctested right?

lrennels: I think you’re right we should change or remove it since it isn’t merged yet and we’d want these blocks doctested.

lrennels:
@ckingdon95 things pass now!  I think we just need an approving review and we can merge.

src/core/defs.jl
```diff
@@ -353,9 +353,9 @@ end
 """
     parameter_dimensions(obj::AbstractComponentDef, param_name::Symbol)
 
-Returns the names of the dimensions of parameter `param_name`.
+Return the names of the dimensions of parameter `param_name` exposed in the composite 
+component definition indicated by`obj`.
```
ckingdon95: Is `obj` always/only composite definitions? Or can it also just be a component definition?

docs/Project.toml
```diff
@@ -1,5 +1,17 @@
...
+MimiDICE2010 = "4f2c5d70-42a7-11e9-1de8-c370bd984083"
+MimiFUND = "b3ba11de-429f-11e9-29f7-cb478ab96e7c"
+StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
+VegaLite = "112f6efa-9a02-5b7d-90c0-432ed331239a"
 
 [compat]
-Documenter = "~0.21"
+Distributions = "0.21"
+Documenter = "0.23"
+Mimi = "0.9.3"
```
ckingdon95: should this be Mimi 0.9.4?

docs/src/tutorials/tutorial_1.md
```diff
@@ -5,59 +5,70 @@ This tutorial walks through the steps to download, run, and view the output of a
...
+
+Note that these steps should be relatively consistent across models, where a repository for `ModelX` should contain a primary file `ModelX.jl` which exports, at minimum, a function named something like `get_model` or `construct_model` which returns a version of the model, and can allow for model customization within the call.
+
+In the MimiFUND package, the function `get_model` has the signature
+```julia
+get_model(; nsteps = default_nsteps, datadir = default_datadir, params = default_params)
 ```
+Thus there are no required arguments, although the user can input `nsteps` to define the number of timesteps (years in this case) the model runs for, `datadir` to define the location of the input data, and `params`, a dictionary definining the parameters of the model.  For example, if you wish to run only the first 200 timesteps, you may use:
+```jldoctest tutorial1; output = false, filter = r".*"s
```
ckingdon95: why is this `output = false` here? Don't we want to make sure the model runs without error?

docs/src/tutorials/tutorial_1.md
```diff
@@ -67,10 +78,14 @@ m[:ComponentName, :VariableName] # returns the whole array of values
 m[:ComponentName, :VariableName][100] # returns just the 100th value
 
 ```
-Indexing into a model with the name of the component and variable will return an array with values from each timestep. You may index into this array to get one value (as in the second line, which returns just the 100th value). Note that if the requested variable is two-dimensional, then a 2-D array will be returned. For example, try taking a look at the `income` variable of the `socioeconomic` component using the code below:
-```
+Indexing into a model with the name of the component and variable will return an array with values from each timestep. You may index into this array to get one value (as in the second line, which returns just the 100th value). Note that if the requested variable is two-dimensional, then a 2-D array will be returned. For example, try taking a look at the `income` variable of the `socioeconomic` component of FUND using the code below:
+```jldoctest tutorial1; output = false
```
ckingdon95: why false?

docs/src/tutorials/tutorial_1.md
```diff
@@ -79,12 +94,35 @@ You may also get data in the form of a dataframe, which will display the corresp
 getdataframe(m, :ComponentName=>:Variable) # request one variable from one component
 getdataframe(m, :ComponentName=>(:Variable1, :Variable2)) # request multiple variables from the same component
 getdataframe(m, :Component1=>:Var1, :Component2=>:Var2) # request variables from different components
-
 ```
 
 Try doing this for the `income` variable of the `socioeconomic` component using:
-```
+```jldoctest tutorial1; output = false
```
ckingdon95: why false?

docs/src/tutorials/tutorial_1.md
```diff
@@ -98,23 +136,26 @@ Mimi provides support for plotting using [VegaLite](https://github.com/vega/vega
 If you wish to explore the results graphically, use the explorer UI, described [here](https://www.mimiframework.org/Mimi.jl/stable/userguide/#Plotting-and-the-Explorer-UI-1) in Section 5 of the Mimi User Guide.
 
 To explore all variables and parameters of FUND in a dynamic UI app window, use the [`explore`](@ref) function called with the model as the required first argument, and the optional argument of the `title`  The menu on the left hand side will list each element in a label formatted as `component: variable/parameter`.
-```
+```jldoctest tutorial1; output = false, filter = r".*"s
```
ckingdon95: same here

docs/src/tutorials/tutorial_2.md
```diff
@@ -26,67 +24,86 @@ When the original model calls [`set_param!`](@ref), Mimi creates an external par
...
+You have now successfully downloaded MimiDICE2010 to your local machine.
 
 ### Step 2. Run DICE
 
-The next step is to run DICE.  If you wish to first get more aquainted with the model itself, take a look at the provided online documentation.  Now run DICE using the provided API for the package:
+The next step is to run DICE using the provided API for the package:
 
-```
+```jldoctest tutorial2; output = false, filter = r".*"s
 using MimiDICE2010
```
ckingdon95: It seems like all of the tutorial blocks now have `output = false`, does that mean they aren't being tested?

docs/Project.toml
ckingdon95: Or actually wouldn't we need it to be the 0.9.5-dev version? I guess this is a broader question I have about whether or not our documentation is supposed to be compatible with the latest released version, or with master. I _think_ that the version of documentation that lives on https://www.mimiframework.org/Mimi.jl/stable/ only shows the documentation for the latest _released_ version right? So that means we should always have our documentation up to date with master, but it won't actually display until a new release of Mimi happens? 

If all of that is true, then I think the version of Mimi that we want in the Project.toml would be master, but I'm not sure how to specify that

src/core/defs.jl
lrennels: I think it can probably be either

docs/Project.toml
lrennels: Yes, the version of the documentation that lives in the `stable` link only shows documentation for the latest **tagged/released** version of Mimi, so the documentation might as well keep up with master.

davidanthoff: I think you should remove all entries other than `Documenter` itself from here. Documenter should be specified as `~0.23`.

ckingdon95: What does the `~` mean? I think we also need to keep MimiFUND at at least `3.11.9`, so would that also need to be `MimiFUND = "~3.11.9"` here?

davidanthoff: https://julialang.github.io/Pkg.jl/v1/compatibility/#Caret-specifiers-1

So I actually think it doesn't matter whether we use caret or tilde for `Documenter`, we should just specify one of them.

For FUND, if there is a minimal version we require we can just say `MimiFUND = "3.11.9"`. But do we need that? It will just grab the latest version if we don't specify anything, which is what we want, right?

davidanthoff: Ok, but for FUND, I think we should actually _not_ use the tilde operator, otherwise we potentially pin FUND to old versions.

ckingdon95: If it will always grab the most recent version, then why do we need to specify for Documenter?

ckingdon95: Ok, we shouldn't have to specify FUND then. But why do we have to specify for Documenter then?

davidanthoff: They might release a breaking release, and we would want to opt in to get that.

ckingdon95:
> @ckingdon95 the tests are passing here, and I needed to remove the actual `doctest(Mimi)` call from the `make.jl` file to get the build to run. That said, now when doctests fail it shows up in the log but does not cause the build to fail. I'll ask about this on Slack.
> 
> **update:** After some conversation I've solved this for now by doing our `doctest` outside of the `makedocs` call, and then setting the `doctest` keyword arg in that function to `false` so that the build doesn't stall when trying to test docs twice. I've tried a few options and I believe this is the most robust for now.

@lrennels Just want to make sure I'm understanding the status of this: the way it's now set up (having doctest(Mimi) outside of the makedocs call), is it still true that if the doctests fail, that the build won't fail?

davidanthoff:
Are we doing [this](https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Doctesting-as-Part-of-Testing-1)? I think we should, right? We want the doctests to run during a normal test run, right?

lrennels:
There are a few options for how to run the doctests.  The default way is for doctesting to occur within the `makedocs` function, but that does not cause the build to fail unless you set `strict = true`, which is strict on other things that we do not fully adhere to and thus causes failure even if doctests pass (i.e. docstrings for every single function, 95 of which we are missing).

ckingdon95:
Yeah I think we should just move it into our test folder (what David said), then it will use the Test environment instead (we just need to add MimiFUND, MimiDICE, and Documenter to the test environment I think). I can make that change if that's what we think is best!

lrennels:
That's fine by me!

davidanthoff:
I don't think we can delete the `Project.toml` from the `docs` folder entirely, right? Still needed for the docs build stage?

I have to say that I _really_ don't like the whole design around that in Documenter.jl...

lrennels:
@davidanthoff is correct we can't delete the Project.toml we still need it!

ckingdon95:
Ok I'll add it back! Wasn't sure. Two other questions though:
- Now that Documenter needs to be specified in the `[targets] test = [ ]` section of the main Project.toml, how do we specify the version? I added `Documenter = "~0.23"` to the [compat] section, but wasn't sure if that was right since it's not in the regular [deps] list
- When Lisa originally had the doctests running in the docs environment, she had added
`julia --project=docs/ -e 'using Pkg; Pkg.Registry.add(RegistrySpec(url ="https://github.com/mimiframework/MimiRegistry.git"))'` 
to Travis so that MimiFUND and MimiDICE could be used. What is the equivalent thing I need to do so that they can be used during runtests? I think the current Travis build failed because it couldn't find MimiDICE

lrennels:
For your first question, I think I tried to look this up but had trouble finding an answer, I can look again.  For the second, I think you might need to add that line in to the travis and apveyor .yml files?

ckingdon95:
Also can someone explain why we need the Project.toml in the docs folder in order for them to build? 

And when I add it back, would it only need the following right?
```
[deps]
Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
Mimi = "e4e893b0-ee5e-52ea-8111-44b3bdec128c"

[compat]
Documenter = "~0.23"
```
Since it previously had several other packages for running the doctests, but we wouldn't need them anymore?

lrennels:
I'll be honest I don't know, I only said that because I know it was there before I started messing with doctests ... good question. I guess it's just a characteristic of `Documenter.jl` that it's going to need a Project.toml since it won't use the one from Travis or Apveyor?

lrennels:
@davidanthoff I think we need your sage wisdom to clarify the past few questions but I'll look around documentation to see what I can find! Also @ckingdon95 the responses on the documentation channel of Julia slack are incredibly fast, so that's been helpful for me but I think in this case David can probably answer :)

davidanthoff:
So I think for the doc build stage we need the `Project.toml` because the whole docs [job](https://github.com/mimiframework/Mimi.jl/blob/master/.travis.yml#L31) assumes that there is one in the `docs` folder when it tries to build the docs (note how it starts julia with the `--project=docs` flag).

It is kind of frustrating that the `doctest` runs in the testing environment, rather than in the doc environment... But, ok, what can we do :) So I think we probably need to add a section to `.travis.yml` like this:
```
before_script:
  - julia -e 'using Pkg; Pkg.Registry.add(RegistrySpec(url ="https://github.com/mimiframework/MimiRegistry.git"))'
```
I think that should probably be enough so that we can specify `MimiFUND` etc. as a test dependency. BUT, I'm generally not super happy with that setup... I somehow feel that really our tests should work without the need to add another registry. But maybe lets ignore that problem for now and just go ahead.

I would suggest we drop appveyor. Travis now supports testing on Windows, so that should simplify things. I'll open a PR that makes that change.

lrennels:
Ok yea that all makes sense. I think that’s why I originally called the the `doctest ` function within the docs make file, so that it used the doc environment.

ckingdon95:
@davidanthoff given all of that, would you prefer for `doctest(Mimi)` to happen in "docs/make.jl" like Lisa originally had it?

davidanthoff:
No, I still think they should run as part of the tests.

ckingdon95:
@davidanthoff not exactly sure what this current Travis error is but I'm guessing it has to do with adding the Registry? Maybe it prevented it from finding packages that aren't in our registry? not sure
```
3.22s$ julia --color=yes -e "if VERSION < v\"0.7.0-DEV.5183\"; Pkg.clone(pwd()); Pkg.build(\"${JL_PKG}\"); else using Pkg; if VERSION >= v\"1.1.0-rc1\"; Pkg.build(verbose=true); else Pkg.build(); end; end"
  Updating registry at `~/.julia/registries/MimiRegistry`
  Updating git-repo `https://github.com/mimiframework/MimiRegistry.git`
ERROR: Package LightGraphs [093fc24a-ae57-5d10-9952-331d41423f4d] not found in a registry.
Stacktrace:
 [1] pkgerror(::String) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Pkg/src/Types.jl:112
 [2] check_registered(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Pkg/src/Operations.jl:924
 [3] up(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}, ::Pkg.Types.UpgradeLevel) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Pkg/src/Operations.jl:1043
 [4] #up#43(::Pkg.Types.UpgradeLevel, ::Pkg.Types.PackageMode, ::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(Pkg.API.up), ::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Pkg/src/API.jl:167
 [5] #up#38 at ./none:0 [inlined]
 [6] #up at ./none:0 [inlined]
 [7] #instantiate#81(::Nothing, ::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(Pkg.API.instantiate), ::Pkg.Types.Context) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Pkg/src/API.jl:463
 [8] instantiate at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Pkg/src/API.jl:461 [inlined]
 [9] build(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}, ::Bool) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Pkg/src/Operations.jl:710
 [10] #build#73(::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(Pkg.API.build), ::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Pkg/src/API.jl:399
 [11] #build at ./none:0 [inlined]
 [12] #build#72 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Pkg/src/API.jl:380 [inlined]
 [13] #build at ./none:0 [inlined]
 [14] #build#69 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/Pkg/src/API.jl:377 [inlined]
 [15] (::getfield(Pkg.API, Symbol("#kw##build")))(::NamedTuple{(:verbose,),Tuple{Bool}}, ::typeof(Pkg.API.build)) at ./none:0
 [16] top-level scope at none:1
```

.travis.yml
```diff
@@ -25,15 +25,19 @@ branches:
   - /v(\d+)\.(\d+)\.(\d+)/
 env:
   - PYTHON="" MIMI_RUN_DEPENDENCY_TESTS="TRUE"
+before_script:
+  - julia -e 'using Pkg; Pkg.Registry.add(RegistrySpec(url ="https://github.com/mimiframework/MimiRegistry.git"))'
```
davidanthoff: Try to change this to:
```
- julia -e 'using Pkg; Pkg.update(); Pkg.Registry.add(RegistrySpec(url ="https://github.com/mimiframework/MimiRegistry.git"))'
```
The problem here is that if we add a new registry _before_ we've ever used `General`, the latter is not added to the registries at all.

lrennels: I'll try that!  Also just noting the tests won't pass until we remove appveyor, or add this line to the appveyor.yml file somehow!

ckingdon95:
hmm now it's stalling with the error: "No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."

Which is the same thing that's happening on your other PR to run Travis on Windows (except that it's also happening here on mac and linux) @davidanthoff

davidanthoff:
What an endless nightmare, this branch :)

I think the strategy now is to insert `@info` statements between the different `includes` in the `runtests.jl` to figure out where it stalls...

ckingdon95:
@davidanthoff  we already have those @info statements, so we can see that the last one to give output before it stalled was "test_timesteparrays.jl" which is about 3/4 of the way through all the tests in "runtests.jl". Not sure what to do with that info though

ckingdon95:
It did successfully run the doctests I think, since I put that at the top of the "runtests" file, so that's good that that's working in the test environment now. So I guess this is now a different Travis problem?

lrennels:
This branch is my nightmare.  Ok so do we know what the commit is that caused Mac Travis to start failing?  It’s when we changed the .yml file calls to the registry adds etc?

lrennels:
Ah ok I see. This might be off base, but I was running into that exact problem when I ran the doctests once before the documentation ran, and then they were run again in `makedocs`. But I have no idea why.  So maybe something in the doc testing is making the environment stall ... but I have no idea.  Maybe more info statements will help us.

lrennels:
@ckingdon95 @davidanthoff I think there's something going on here with the explorer and broken pipes, which I've seen before.  When we run `doctest(Mimi)`, this includes running several calls to `explore`.  Then in the test suite, `test_timesteparrays.jl` is the first test to call `explore`. 

Trying this locally, if I run `doctest(Mimi)` and let all of the tests run, and then close all of the windows and external Julia instance etc. for `explore`, and then try to run the original testing suite I get an error at `test_timesteparrays.jl` that is familiar:

```
Got exception outside of a @test
  LoadError: IOError: write: broken pipe (EPIPE)
  Stacktrace:
   [1] wait(::Task) at ./task.jl:198
   [2] fetch at ./task.jl:211 [inlined]
   [3] req_response(::Electron._Application{Electron.Window}, ::Dict{String,Any}) at /Users/lisarennels/.julia/packages/Electron/wL84D/src/Electron.jl:226
   [4] Electron.Window(::Electron._Application{Electron.Window}, ::Dict{String,Any}) at /Users/lisarennels/.julia/packages/Electron/wL84D/src/Electron.jl:291
   [5] #Window#15 at /Users/lisarennels/.julia/packages/Electron/wL84D/src/Electron.jl:309 [inlined]
   [6] (::getfield(Core, Symbol("#kw#Type")))(::NamedTuple{(:options,),Tuple{Dict{String,Any}}}, ::Type{Electron.Window}, ::Electron._Application{Electron.Window}, ::URIParser.URI) at ./none:0
   [7] #explore#224(::String, ::typeof(explore), ::Model) at /Users/lisarennels/.julia/dev/Mimi/src/explorer/explore.jl:34
```

While I do not see this error in the testing log, it's possible it's related.  One solution might be to remove tests of `explore` from the `doctests`, another might be to move the `doctest` call to below the testing suite but I'm not sure that would help.

davidanthoff:
Can we try the following:
1. Remove the call to `explore` from `test_timesteparrays.jl`, that really shouldn't be there, right?
2. Mark all code examples in the documentation that call `explore` as not a doctest, so that they don't run. We can still try to enable that later, but maybe that helps for now.

I tried to look through Electron.jl to figure out how that error that @lrennels reported might come about, but I don't understand. @lrennels could you actually open an issue over at Electron.jl with a replication steps for that pipe error, just so we don't lose track of it?

lrennels:
Yep I can certainly do both of those things! Per #1, I can move the specific test we are doing to `test_explorer_model.jl`. The goal is testing to make sure we can call `explore` properly on a model for which `time` is not the first dimension, so we will want to keep that test somewhere.

davidanthoff:
So I think many of the structural problems are now ok, but now we are running into output comparison tests that depend on a digit very far to the left, and we get slight differences there on different platforms, so the tests fail... 

I guess we could a) for now just not compare outputs for these specific tests, just so that we can merge this and then fix remaining problems in smaller follow up PRs, or b) try to output the results with less digits so that we don't run into these problems. But that of course is tricky because it would make our example code much more complicated with something that is really only needed for the tests....

@lrennels  and @ckingdon95 in any case, can you two take it from here? Let me know if you run into more of the infrastructure problems again, happy to help with those!

Issue #567, opened 2019-10-07
decide whether or not to allow first/last keywords


ckingdon95:
Currently on master you can't have components of different time lengths. The use of `first` and `last` keywords in `add_comp!` was disabled in order to get composite components working. I think before v1.0.0 we should decide if we are going to try to get this working, or if not, then I think we should remove the keywords from the `add_comp!` function all together.

ckingdon95:
If we won't allow this, then I think there are also some `getindex` functions we could remove, that were implemented for the case where the `TIMES` of the timestep don't match the `TIMES` of the timestep array, which would no longer be needed

davidanthoff:
For now disable last and first keyword for 1.0.0.

ckingdon95:
delete keywords from add_comp function, and remove from documentation

Issue #422, opened 2019-03-08
What behavior do we want when updating model time dimension?


rjplevin:
In the test file `test_dimensions.jl`, we set the first/last period of a component to 2005 and 2105. We then call `set_dimension!(m, :time, 2010:2050)` and test that the component's first/last have been updated.

This strikes me as incorrect behavior. If the component can run outside the range the model is being run for, I don't think we should change the component, though this raises questions about component initialization if it's started after it's indicated first period.

Bottom line is that I'm not sure what the correct behavior here should be. Both updating it and not updating it can result in confusion and violation of the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Seeking feedback...

ckingdon95:
I think I currently rely on this updating of time periods in components. I think this makes sense for now because a component can't run longer than the model (so if you're shrinking the time span of the model, then the component has to run on that too). I realize this is not a strong argument, but would vote for keeping this functionality as is.

rjplevin:
Consider first that we have potentially 3 actors in the mix:
* A composite component developer
* A model (or composite component) developer using others' composite components
* An end user using the above

We need a way to communicate assumptions among these parties about how to use the components "correctly" with respect to time. A user should be able to run a component in an "off label" mode, but they should know they're doing so.

There are two main cases:

1. A model or composite component contains a leaf component with a later start or earlier end than the current model timespan, i.e., the component timespan is binding. 

   The component won't run until the model reaches the component's `first` period, which seems ok. Do we run the components' `init` function when the model starts or just before the component starts? 

   The component will stop running before the model does, which may result in `missing` values. I can imagine cases where stopping the component early is fine, and cases in which it breaks the model. We don't want this to appear to be a random outcome from the user perspective.

2. A composite component contains a leaf component with a earlier start or later end than the model timespan, i.e., the model timespan is binding. I think this is the use case @ckingdon95 described.

   A potential problem with starting the model later than the component was designed for is that "first period" initialization steps might be skipped. 
   * Do we need a way to indicate that a component is not valid unless it starts at a certain timestep?
   * Do we need guidelines for component design and use of the `init` function to create "time-flexible" components? Maybe this is just a matter of documentation.

   A component with a `last` period beyond the model isn't a problem; it just won't be run as far as it is designed to.

Having never built a real Mimi model, I don't have the best feel for the model development process and why/when one would be altering time horizons.

rjplevin:
[...]

PR #578, opened 2019-10-15
Remove first and last keyword arguments

ckingdon95:
#567

### DIFF
--- a/src/core/defs.jl
+++ b/src/core/defs.jl
@@ -754,55 +754,39 @@ function import_params!(comp::AbstractComponentDef)
 end
 
 """
-    add_comp!(obj::AbstractCompositeComponentDef, comp_def::AbstractComponentDef,
-              comp_name::Symbol=comp_def.comp_id.comp_name; rename=nothing,
-              first=nothing, last=nothing, before=nothing, after=nothing)
+    add_comp!(
+        obj::AbstractCompositeComponentDef, 
+        comp_def::AbstractComponentDef,
+        comp_name::Symbol=comp_def.comp_id.comp_name; 
+        before::NothingSymbol=nothing, 
+        after::NothingSymbol=nothing,
+        rename::NothingPairList=nothing
+    )
 
 Add the component `comp_def` to the composite component indicated by `obj`. The component is 
 added at the end of the list unless one of the keywords `before` or `after` is specified.
 Note that a copy of `comp_id` is made in the composite and assigned the give name. The optional
 argument `rename` can be a list of pairs indicating `original_name => imported_name`.
-
-Note: `first` and `last` keywords are currently disabled.
 """
-function add_comp!(obj::AbstractCompositeComponentDef, comp_def::AbstractComponentDef,
-                   comp_name::Symbol=comp_def.comp_id.comp_name;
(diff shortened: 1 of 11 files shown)


src/core/defs.jl
```diff
@@ -754,55 +754,39 @@ function import_params!(comp::AbstractComponentDef)
...
-
-        if first !== nothing && first < time_index[1]
-            error("Cannot add component $comp_name with first time before first of model's time index range.")
-        end
-
-        if last !== nothing && last > time_index[end]
-            error("Cannot add component $comp_name with last time after end of model's time index range.")
-        end
-
-        if before !== nothing && after !== nothing
```
rjplevin: I think the `before` and `after` stuff should remain, right?

ckingdon95: I moved it outside of the `if` statement, since I think it should always be checked, not just when there is a time dimension? (although I think all components have a time dimension, so I think it was always being checked)

ckingdon95:
@rjplevin I removed the use of `first` and `last` keywords in the various `add_comp!` and `replace!` methods. Do you think I should also remove the `first` and `last` fields from the `ComponentDef`, `ComponentInstance`, and `ModelComponentDef` classes for this pull request? example:

```julia
@class mutable ComponentDef <: NamedObj begin
    comp_id::Union{Nothing, ComponentId}    # allow anonynous top-level (composite) 
    ComponentDefs (must be referenced by a ModelDef)
    comp_path::Union{Nothing, ComponentPath}
    dim_dict::OrderedDict{Symbol, Union{Nothing, Dimension}}
    namespace::OrderedDict{Symbol, Any}
    first::Union{Nothing, Int}
    last::Union{Nothing, Int}
    is_uniform::Bool

    ...

end

**Comment by rjplevin** [comment]
Agreed, no need to leave artifacts behind.

**Comment by ckingdon95** [comment]
@rjplevin It turns out that deleting the `first` and `last` fields from those three classes is a bit more involved, since there are also these `first_period` and `last_period` methods for each type:
```julia
first_period(obj::AbstractComponentDef) = obj.first
first_period(obj::AbstractComponentInstance) = obj.first
first_period(root::AbstractCompositeComponentDef, comp::AbstractComponentDef) = @or(first_period(comp), first_period(root))
```
These methods get used all throughout the code when setting parameters, making connections, and allocating timestep arrays (even though for now the first period is always just the first year of the time index, since the first/last functionality was disabled). I would have to delete and work around all the calls to `first_period` in these places, but I'm wondering if it would be better to leave this untouched for v1.0.0, and then deal with them later when we might re-implement the first/last capabilities in a different way, as we were discussing at the end of this issue: https://github.com/mimiframework/Mimi.jl/issues/422

Thoughts?

rjplevin:
I’m out running errands today, so slow in responding. 

I think leaving it is reasonable for now.

Cheers,
Rich

src/core/defs.jl
ckingdon95: @rjplevin is this okay?

rjplevin: Yes, I think so. (Honestly, I haven't looked very closely, but I trust you've thought it through!)

Issue #538, opened 2019-08-24
d.time should return AbstractTimestep things


davidanthoff:
I think that `d.time` inside `run_timestep` should return things that one can use to index into variables and parameters directly. Right now it seems to return `Int`s instead, which we should actively discourage for indexing use.

Ideally things like `first(d.time)` and `last(d.time)` would just work. And then we should try to get something like `first(d.time):last(d.time)` to work to get a range that includes every timestep.

ckingdon95:
I'll plan to work on this after we finish #551. After that is merged, we will have multiple types of `Timestep` things that can be used as indexes, so we'll have to decide which kind `d.time` should return

ckingdon95:
We decided that this should not be either of the new `AbstractTimesteps` (`TimestepValue` or `TimestepIndex`). This should return either a `VariableTimestep` or a `FixedTimestep`, whichever it is to match the `t` being passed to the `run_timestep` function. @davidanthoff, I think I will just implement this as an array, unless you choose otherwise! currently d.dimname returns vectors

rjplevin:
Are the new indexers `AbstractTimestep`s? This seems odd to me, since they have to be resolved against an (old) `AbstractTimestep`. I think we need a new abstract type.

ckingdon95:
Oh I'm not sure, that implementation isn't finished yet. But maybe you should let Lisa know here: https://github.com/mimiframework/Mimi.jl/pull/551

davidanthoff:
Shouldn't all four of these inherit from `AbstractTimestep`?

I think returning an array is fine for now. We can later still decide to change that to something more efficient.

rjplevin:
> Shouldn't all four of these inherit from `AbstractTimestep`?

There are methods that make sense for `FixedTimestep` and `VariableTimestep` that cannot work with the new `TimestepValue` or `TimestepIndex`, which are not themselves timesteps (in the sense of the prior two), but indices into the time dimension.

What exactly is gained by unifying these under `AbstractTimestep`? The usage of the two pairs is necessarily different.

rjplevin:
Maybe we should rename `FixedTimestep` and `VariableTimestep` to be `FixedTimeDimension` and `VariableTimeDimension`?

davidanthoff:
I'm not sure :) Maybe it all doesn't matter that much in any case, actually? The only reason to have an abstract type at all is if we use that for dispatch somewhere, right?

rjplevin:
[...]

PR #581, opened 2019-10-16
Make d.time return AbstractTimesteps

ckingdon95:
Issue #538 

The `DimValueDict` previously held vectors of Ints for each dimension (keyed by a Symbol for the name of the dimension). Now, I've changed it so that the `:time` entry in this dictionary holds a vector of `AbstractTimestep`'s, so that calls to `d.time` in the `run_timestep` functions return timesteps that can be used for indexing into parameters and variables. There is only one place in the code where a `DimValueDict` object is constructed, and that is in the `run` function so that it can be passed to the `run_timestep` function.

Questions:
1. What I've implemented here was to modify the constructor of a `DimValueDict` to also take in the `Clock` object from which the `:time` vector of timesteps is created and stored in the dictionary (instead of a vector of Ints). Alternatively, I could have left the `DimValueDict` constructor untouched, so that it still would only hold vectors of Ints for each dimension, and then instead I would have to replace the `:time` entry of the `dim_value_dict` within the `run` function before it gets passed to `run_timestep` (there was a comment in the `run` function that you'll see I deleted, that I think was considering this alternative approach). I think I prefer the way I have implemented it here though, since the `DimValueDict` type only gets used for this one purpose, and also because we would still have to modify the type of the `dict` field to be `Dict{Symbol, Union{Vector{Int}, Vector{AbstractTimestep}}}` so that we could modify the `:time` entry afterward. But I am open to doing it the other way if you think that would be better!
2. Are you okay with this `Base.collect(clock::Clock)` function I implemented here to get the vector of timesteps that go into the dictionary, or should I call it something else?
3. How I've implemented this is to have `d.time` return the vector of timesteps that the Clock was going to run for. I think this is what we want users to have access to during `run_timetsep`, but I could be wrong, and maybe we want users to have access to the full time dimension, even if the model is being run for a shorter amount of time. I think @davidanthoff should give input here.
4. I added a minimal test to the "test_dimensions.jl" file, let me know if you think there is a better place to put these new tests!

### DIFF
--- a/src/core/instances.jl
+++ b/src/core/instances.jl
@@ -283,13 +283,14 @@ function Base.run(mi::ModelInstance, ntimesteps::Int=typemax(Int),
         time_keys = time_keys[1:ntimesteps]
     end
 
-    # TBD: Pass this, but substitute t from above?
-    dim_val_dict = DimValueDict(dim_dict(mi.md))
+    clock = Clock(time_keys)
+
+    # Get the dimensions dictionary
+    dim_val_dict = DimValueDict(dim_dict(mi.md), clock)
 
     # recursively initializes all components
     init(mi, dim_val_dict)
 
-    clock = Clock(time_keys)
     while ! finished(clock)
         run_timestep(mi, clock, dim_val_dict)
         advance(clock)
--- a/src/core/time.jl
+++ b/src/core/time.jl
@@ -172,3 +172,13 @@ function Base.reset(c::Clock)
 	c.ts = c.ts - (c.ts.t - 1)
 	nothing
 end
+
+function Base.collect(c::Clock)
+	c = deepcopy(c)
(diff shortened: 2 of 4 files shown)


ckingdon95:
@rjplevin when you have time, I'd like your feedback on these questions! thanks!

rjplevin:
1. I think modifying the constructor to provide exactly what's needed in `run_timestep` is the right approach. 

2. I believe the "standard" use of `collect` is to, well, collect values from an iterator. I would avoid confusion and just call it `timesteps(c::Clock)`.

3. I agree; David should weigh in.

4. Sure, however you decide to organize the tests is fine.

lrennels:
@ckingdon95 we want to merge this with/before my `timestepindex` branch right?

PR #582, opened 2019-10-17
Update references.md

lrennels:
Remove unexported functions from reference list

### DIFF
--- a/src/core/defs.jl
+++ b/src/core/defs.jl
@@ -350,6 +350,11 @@ function parameter_unit(obj::AbstractComponentDef, param_name::Symbol)
     return unit(param)
 end
 
+"""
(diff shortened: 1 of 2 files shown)


src/core/defs.jl
```diff
@@ -567,6 +577,11 @@ function variable_dimensions(obj::AbstractCompositeComponentDef, comp_path::Comp
     return dim_names(var)
 end
```
ckingdon95: Should you also add the same docstring to this one?

lrennels: Yea I can add it to the two other methods.  I wasn't quite clear on what a `ComponentPath` type was etc. but can write a docstring for that without a problem.

lrennels: I took a look at other doc strings for inspiration, this is done!

PR #583, opened 2019-10-24
Add Confidence Intervals to Sobol Indices

lrennels:
This PR adds two keyword arguments, `num_resamples::Int = 10_000` and `conf_level::Number = 0.95` to the `analyze` function currently only used in Sobol analysis.  These args will be passed to `GlobalSensitivityAnalysis.analyze` and direct the calculation of confidence intervals around the total, first, and second order indices.

This PR is in combination with more significant work on the `GlobalSensitivityAnalysis.jl` side of things, which needs a bit more testing but will then be included in the next tagged version this upcoming week.

https://github.com/lrennels/GlobalSensitivityAnalysis.jl/pull/21

(FYI @ckingdon95 @davidanthoff)


lrennels:
@ckingdon95 would you mind approving this so we can get it into master now? I'm trying to do some work with it but it's getting complicated with all the branches and I'm going to end up holding off until we get all the models up to speed etc.

PR #588, opened 2019-11-15
Add docstrings

lrennels:


src/core/defs.jl
```diff
@@ -578,7 +578,10 @@ unit(obj::ParameterDefReference) = parameter(obj).unit
     variable_dimensions(obj::AbstractCompositeComponentDef, comp_path::ComponentPath, var_name::Symbol)
 
 Return the names of the dimensions of variable `var_name` exposed in the composite 
-component definition indicated by`obj` along the component path `comp_path`.
+component definition indicated by`obj` along the component path `comp_path`. The 
+`comp_path`is of type `Mimi.ComponentPath` with the single arugment being a named
```
ckingdon95: "argument" typo

lrennels: @ckingdon95 fixed!

ckingdon95:
Do we need this PR now? it looks like these things were also added in the other one?

lrennels:
Ah true I should edit them in that one too.  The other one was stalling for so long that I wanted to get this part done, since it was breaking the docs build.

ckingdon95:
I meant the "variable_dimensions" one. I think that one can be merged now (it's passing) and then we don't need this one? Cause you also added these docstrings to that one?

lrennels:
Yep you're right I just changed that, so the only thing that this one does now is add docstrings for `TimestepValue` and `TimestepIndex` which we do need, because they are in the reference.md file but missing docstrings.

lrennels:
Oh they're both in that PR!  Ok yes you're right let me fix that right now.

PR #589, opened 2019-11-15
Add variable_dimensions methods

lrennels:
@jrising

### DIFF
--- a/src/core/defs.jl
+++ b/src/core/defs.jl
@@ -579,8 +579,8 @@ unit(obj::ParameterDefReference) = parameter(obj).unit
 
 Return the names of the dimensions of variable `var_name` exposed in the composite 
 component definition indicated by`obj` along the component path `comp_path`. The 
-`comp_path`is of type `Mimi.ComponentPath` with the single arugment being a named
-tuple of symbols describing the relative (to a composite) or absolute (relative to ModelDef) 
+`comp_path`is of type `Mimi.ComponentPath` with the single field being an NTuple 
+of symbols describing the relative (to a composite) or absolute (relative to ModelDef) 
 path through composite nodes to specific composite or leaf node.
 """
 function variable_dimensions(obj::AbstractCompositeComponentDef, comp_path::ComponentPath, var_name::Symbol)
@@ -607,12 +607,7 @@ component definition indicated by`obj` along the component path `comp_path`. The
 `comp_path` is a tuple of symbols describing the relative (to a composite) or 
 absolute (relative to ModelDef) path through composite nodes to specific composite or leaf node.
 """
-function variable_dimensions(obj::AbstractCompositeComponentDef, comp_path::Tuple{N, Symbol}, var_name::Symbol) where N
-    return variable_dimensions(obj, Mimi.ComponentPath((comp_path)), var_name)
-end
-
-# special case for a Tuple with a single element, shouldn't really be used but could come up
-function variable_dimensions(obj::AbstractCompositeComponentDef, comp_path::Tuple{Symbol}, var_name::Symbol) where N
+function variable_dimensions(obj::AbstractCompositeComponentDef, comp_path::NTuple{N, Symbol}, var_name::Symbol) where N
     return variable_dimensions(obj, Mimi.ComponentPath((comp_path)), var_name)
 end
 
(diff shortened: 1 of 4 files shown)


src/core/defs.jl
```diff
@@ -578,13 +578,44 @@ unit(obj::ParameterDefReference) = parameter(obj).unit
     variable_dimensions(obj::AbstractCompositeComponentDef, comp_path::ComponentPath, var_name::Symbol)
 
 Return the names of the dimensions of variable `var_name` exposed in the composite 
-component definition indicated by`obj` along the component path `comp_path`.
+component definition indicated by`obj` along the component path `comp_path`. The 
+`comp_path`is of type `Mimi.ComponentPath` with the single arugment being a named
+tuple of symbols describing the relative (to a composite) or absolute (relative to ModelDef) 
+path through composite nodes to specific composite or leaf node.
```
ckingdon95: saying "a named tuple" confused me at first, cause it's actually just a Tuple right?

src/core/defs.jl
```diff
@@ -578,13 +578,44 @@ unit(obj::ParameterDefReference) = parameter(obj).unit
...
+absolute (relative to ModelDef) path through composite nodes to specific composite or leaf node.
+"""
+function variable_dimensions(obj::AbstractCompositeComponentDef, comp_path::Tuple{N, Symbol}, var_name::Symbol) where N
+    return variable_dimensions(obj, Mimi.ComponentPath((comp_path)), var_name)
+end
+
+# special case for a Tuple with a single element, shouldn't really be used but could come up
+function variable_dimensions(obj::AbstractCompositeComponentDef, comp_path::Tuple{Symbol}, var_name::Symbol) where N
+    return variable_dimensions(obj, Mimi.ComponentPath((comp_path)), var_name)
+end
```
ckingdon95: Sorry I'm just wrapping my head around this, why do we need the second method? wouldn't a single element tuple stille be handled by the method defined before? If not though, I think the "where N" should be removed

src/core/defs.jl
lrennels: Hm I could reword this slightly, but I had copied it from @rjplevin's response on the forum. For the `ComponentPath` type, the single **field**, not **argument**, is of type `NTuple` (named tuple) which is not the same as `Tuple`.
```
# ComponentPath identifies the path through multiple composites to a leaf comp.
struct ComponentPath <: MimiStruct
    names::NTuple{N, Symbol} where N
end
```
That said, the constructor allows you to input a `Tuple` as the argument, which it will convert to `NTuple`, as shown by something like:
```
function variable_dimensions(obj::AbstractCompositeComponentDef, comp::Symbol, var_name::Symbol)
    return variable_dimensions(obj, Mimi.ComponentPath((comp,)), var_name)
end
```

lrennels: The doctoring could instead read something like:
"Return the names of the dimensions of variable `var_name` exposed in the composite component definition indicated by`obj` along the component path `comp_path`. The  `comp_path`is of type `Mimi.ComponentPath` with a single field, a Named Tuple of symbols describing the relative (to a composite) or absolute (relative to ModelDef) path through composite nodes to specific composite or leaf node. The constructor for a `ComponentPath` can, however, take a regular `Tuple`. 

@rjplevin what do you think the right wording is here?

src/core/defs.jl
lrennels: I thought so too, but Julia seems to internally simplify the case where there is only one element in the Tuple so it does not get picked up by the first method.  It seems to me unlikely that someone would pass `(comp_name,)` instead of just `comp_name`, but if they did we need to his method for that case. You're right I can remove the `where N`

rjplevin: ComponentPath holds an NTuple, not a NamedTuple.

rjplevin: The `where N` in the `Tuple{N, Symbol}` allows it to be a Tuple of any number of Symbols.

lrennels: Oh ok, for some reason without the second method it didn't seem to be picking up a call with a single element `Tuple` but I can try again.

src/core/defs.jl
lrennels: I think @rjplevin's comment below clears this up for me so I can fix it, as you said @ckingdon95 it's not a `NamedTuple.

lrennels:
@ckingdon95 I've addressed the requested changes thank you!

src/core/types/time.jl
```diff
@@ -31,6 +38,12 @@ struct TimestepValue{T}
     end
 end
   
+"""
+     TimestepValue
```
ckingdon95: Sorry didn't catch this till I saw you correct it on the other PR, but this one should be TimestepIndex

lrennels: @ckingdon95 thanks just fixed this, so you're right this encompasses both PRs ... I'll close the docstrings one and wait for tests to pass on this one.  If it looks good now you can approve and we'll be good to go sorry for confusion!

PR #623, opened 2019-12-03
Fix failing tests

lrennels:
reminder to myself:  get all tests passing without doctesting `explore` and `plot` and then try to put those back in in another PR...


davidanthoff:
Joy, tests fail only on 32 bit systems, apparently :)

lrennels:
Ugh, it's all these subtle differences in the output of doctests ... I'll work on it.

davidanthoff:
This is sucking up a lot of time, though... I wish there was some way to run doctests, but only run them, not check their output. Or maybe that can be done? I think that would be good enough for us?

lrennels:
I believe what Cora and I concluded is that we can’t use a flag, but most of the time can use a regular expression and the `filter` keyword arg to filter out all output and thus ignore output.  We don’t do this in all cases, but that is probably going to be my solution now ...

davidanthoff:
Might https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Fixing-Outdated-Doctests-1 help somehow?

davidanthoff:
You probably saw this already: https://github.com/JuliaDocs/Documenter.jl/issues/452#issuecomment-389761169

lrennels:
I haven't tried out the fixing outdated doctests yet, but yes we use the filter suggested in the second link to compare against nothing but still check if the test fails. I think this branch might be set now, but not sure because it's failing before it gets to the doctests due to some compat error.

I am curious about the outdated doctests.  I think one issue is sometimes the output is _slightly_ different on different machines and Julia versions, so sometimes it'll pass on my machine but fail one of the tests ... which is why I punted on some and just filtered the output so it's ignored.

davidanthoff:
Alright, I think we are back to failing doctests?

lrennels:
@davidanthoff I'm getting some frustrating and unpredictable stalling of tests that doesn't seem to be related to our stuff at all ... going to give it a day or two and see if it fixes. All of our tests should run fine, so this will get master passing again and allow us to properly test other PRs.

davidanthoff:
Wait, it all worked? Hurray!!! I thought this saga would never end ;)

lrennels:
Well ... I thought it worked because it let me merge but now I'm worried master isn't going to work because I think two were still pending.  Cross your fingers.

lrennels:
The saga may continue.  But I have no clue what is going on if it doesn't work, all the tests pass and I removed all the doctesting of `plot` and `explore` for now to keep things as clean as I could.

PR #624, opened 2019-12-03
Bump compat for Electron

lrennels:
This pull request changes the compat entry for the Electron package from 1.1 to 1.1, 2.0.

This keeps the compat entries for earlier versions.


lrennels:
Warnings thrown:
```
(node:66353) Electron: Loading non context-aware native modules in the renderer process is deprecated and will stop working at some point in the future, please see https://github.com/electron/electron/issues/18397 for more information
```

Issue #473, opened 2019-05-07
Component Parameter Type Parameterization 


lrennels:
Continuation of #469 discussion, deciding the best way to type parameterize the parameters of a component, with options such as:

1. `param::Array{String, 1} = Parameter(index=[time])`

2. `paramName::String = Parameter(index=[time])`

3.  Avoid overloading Julia syntax and add a keyword to our Parameter definition, e.g.,
`paramName = Parameter(index=[time], type=String)`

davidanthoff:
So we go with option 3, but call the parameter `eltype`.

rjplevin:
Do we want to support syntax option 2 for scalar values? That is, do we allow:

`paramName::Int = Parameter(default=42)`

or require:

`paramName = Parameter(eltype=Int)`

We could also keep both and only error when using `::Type` when indexes are present.

davidanthoff:
I think if we move things into the `Parameter` constructor, we should probably move everything in there?

But one question is whether we should use `eltype` for the scalar case... Does that make sense? Doesn't seem too bad...

ckingdon95:
I thought we were still going to support the (2) syntax, just not for using it to define the `eltype` of an array. But that the following would be okay:
```
paramName::Int = Parameter( )    # paramName would be an integer
paramName::String = Parameter( )    # paramName would be a string
paramName::Matrix{Float64} = Parameter( )    # paramName would be a matrix
```
But this would error, since the index implies that it needs to be an array:
```
paramName::String = Parameter(index=[time])    # not allowed
```

davidanthoff:
Well, should we? I'm unsure... I almost feel that if we do away with the `::Foo` syntax, we should do away with it entirely.

Btw., another option would be `p1 = Parameter{Int}(...)`... The benefit of this would be that we don't need to use a keyword arg like `eltype` that might be wrong for the scalar case.

rjplevin:
> `p1 = Parameter{Int}(...)`
I like that, though it's somewhat more complicated for less-Julia-expert modelers.

ckingdon95:
(I'm just writing this down now so we don't forget what we agreed on. I'm going to start trying to implement soon)

We will implement:
```
p1 = Parameter{Int}()    # ScalarModelParameter that is an Int
p2 = Parameter{Int}(index = [regions])    # ArrayModelParameter whose eltype is Int

p3 = Parameter{Matrix{Float64}}()    # ScalarModelParameter that is a Matrix
```
[...]

PR #638, opened 2019-12-10
New syntax in defcomp for datum type parameterization

ckingdon95:
[Some of the CI tests on this PR will still fail, and we'll need to wait for the other PR that fixes them before we can merge this one]

This PR implements the new syntax for datum type parameterization in `@defcomp`, and errors when a user tries to use the old syntax. Do we want it to just be a deprecation warning instead of an error?

Addresses issue #473


ckingdon95:
@davidanthoff @lrennels 

The tests for this PR are failing because it changes functionality that FUND relies on. (This changes the syntax from `par1::Int = Parameter()` to `par1 = Parameter{Int}()` in `@defcomp`). One of our tutorials uses FUND, so the doctests fail with this change. What would the proper work flow for this be?

We had  removed the model dependency tests out of the normal tests required to merge PRs for exactly this reason, so that we can actually make breaking changes to Mimi master that might break the dependent models, with the intention that we would clean up the models once we are ready to release a new version of Mimi. I guess we forgot that our tutorials use the models as well.

lrennels:
Ah interesting yea it's another one of these circular problems.  I guess this should be fairly rare since it will happen only with breaking changes, but there are a few of those for v1.0.0. 

We could remove the doctest for now, since I don't think we want to allow failing tests to be merged.  We could also move doctests outside of the main testing suite, but that seems a bit dangerous to me because we'd have to make sure to double check those whenever we push to master and I'd prefer them be in the main testing suite. 

I assume we'll change FUND quickly too, since the testing of our dependencies will break, so it's a temporary issue, @davidanthoff do you have a preference for how we handle this?

davidanthoff:
Can we change the PR such that the old syntax is also still supported, with a deprecation warning?

ckingdon95:
> Can we change the PR such that the old syntax is also still supported, with a deprecation warning?

Yes, that would work. I can make it just a warning for now, and then change it to breaking for v1.0.

ckingdon95:
@lrennels you are on it! I just went to add this to the issue with the list of deprecations-to-errors for v1.0 and you beat me to it! thanks!

ckingdon95:
@lrennels good catch! I did actually miss one spot in the docs, I just fixed that. This makes me realize, though, that we never actually fully explain the different type parameterization options in the docs, and we probably should. I'll try to add something later today, probably just to the user guide!

ckingdon95:
@lrennels I just added a new section under Advanced Topics in the userguide, but I feel like I didn't do a great job explaining it. If you have any suggestions for better wording or examples let me know!

PR #642, opened 2019-12-19
Close all applications after testing

lrennels:


lrennels:
@ckingdon95 @davidanthoff HOORAY closing Electron worked!  We can add more doctests later but I think for now we should merge this even though I sort of limited the use of doctests.

PR #648, opened 2020-01-25
Patch v0.9.4 bug in add_connector_comps!

lrennels:
This branch is off of v0.9.4 and patches the bug in the `add_connector_comps` function.  It should not be merged but can be used while we do redesign work around components with different lengths.

### DIFF
--- a/src/core/connections.jl
+++ b/src/core/connections.jl
@@ -465,7 +465,9 @@ end
 
 
 function add_connector_comps(md::ModelDef)
-     conns = md.internal_param_conns        # we modify this, so we don't use functional API
+    conns = md.internal_param_conns        # we modify this, so we don't use functional API
+
+    i = 1 # counter to track the number of connector comps added 
 
     for comp_def in compdefs(md)
         comp_name = name(comp_def)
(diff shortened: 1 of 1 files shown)


ckingdon95:
@lrennels great! just ran this with Frank's code and it works! it feels a little tenuous, because the numbered names (:ConnectorComp1, :ConnectorComp2) get used again in the build function, inside of a loop over the `md.backup` values, which assumes that the values are in the same order as the original naming of the connectorcomps:
https://github.com/mimiframework/Mimi.jl/blob/798275f2c5edec09368d74d8a68459e82b5fd101/src/core/build.jl#L136-L144
which, for now, works, because the "backups" get added in the same order that the original numbering of the names happens, but just something for us to keep in mind that could cause a bug later if any of the ordering changes!

ckingdon95:
So I guess the question now is if we should just tell Frank to use this branch, `conncomps`, which branches off of v0.9.4, or if we want to make it an official version, v0.9.5, since Frank is going to be using it as a version for a paper. Either way it would still be publically available and everything. @davidanthoff which do you prefer?

davidanthoff:
I think we should do a v0.9.5. I can then also update some of the bounds on some of the packages that are being used, which would help a lot with package resolution. I'll prep the release will merge this here into that branch.

lrennels:
I can't seem to get the PR to request a merge into `v0.9.4` instead of `master` ... but that's what we want!

Issue #646, opened 2020-01-24
Fix add_connector_comps! bug


lrennels:
There is a bug in `add_connector_comps` such that if there are needed connector components in more than one of the `comp_def in compdefs(obj)`, we will try to add `ConnectorComp1` twice and trigger an Error.  We need to track how many total have been added outside of the loop over all compdefs. A toy example to show this problem (on v0.9.4) is below from https://forum.mimiframework.org/t/error-in-coupling-different-timestep-length-components/111

This all relates to the bigger issue of adding `first` and `last` keywords #573 
 
```
using Mimi

# A longer CH4 cycle component that runs for 20 periods.
@defcomp ch4_cycle begin
	ch4_emiss = Parameter(index=[time])
	ch4_conc  = Variable(index=[time])

	function run_timestep(p, v, d, t)
    	if is_first(t)
    		v.ch4_conc[t] = 720.0
    	else
    		v.ch4_conc[t] = v.ch4_conc[t-1] + 0.5 * p.ch4_emiss[t]
    	end
    end
end

# A longer CO2 cycle component that runs for 20 periods.
@defcomp co2_cycle begin
	co2_emiss = Parameter(index=[time])
	co2_conc  = Variable(index=[time])

	function run_timestep(p, v, d, t)
    	if is_first(t)
    		v.co2_conc[t] = 278.0
    	else
    		v.co2_conc[t] = v.co2_conc[t-1] + 0.1 * p.co2_emiss[t]
    	end
    end
end

# A shorter emissions component that switches on in period 15.
@defcomp emissions begin
	gdp	      = Parameter(index=[time])
	co2_emiss = Variable(index=[time])
	ch4_emiss  = Variable(index=[time])

	function run_timestep(p, v, d, t)
		v.co2_emiss[t] = 0.25 * p.gdp[t]
		v.ch4_emiss[t] = 0.1 * p.gdp[t]
    end
end


# Create the model for 20 time steps.
m = Model()
set_dimension!(m, :time, 20)

# Add components (emissions turn on in period 15).
add_comp!(m, emissions; first=15)
add_comp!(m, co2_cycle)
add_comp!(m, ch4_cycle)

# Set exogenous gdp scenario for periods 15-20 in emissions component.
set_param!(m, :emissions, :gdp, ones(6) .* 100)

# Create component connections (with backup CO2 and CH4 emissions data for periods 1-15)
backup_ch4_data = ones(20) .* 15
[...]

PR #649, opened 2020-01-25
Fix bug in add_connector_comps on master

lrennels:
#646 This PR fixes a bug in the `add_connector_comps!` function on master, where we have currently disabled functionality surrounding components with different lengths ... but this is a bug we will want to track and fix anyways unless we remove the function all together.  I'd suggest merging this fix for now.


ckingdon95:
@lrennels agreed we should just merge this even though it can't fully be used in this state, it's still a bug!

PR #654, opened 2020-02-11
Fix parameter type bug for non-time arrays

ckingdon95:
@jrising found this bug, more info on this forum post:
https://forum.mimiframework.org/t/array-parameter-setting-fails-with-int64-parameter-type/113

I added a line to the test file which now passes, but represents what previously would have errored.

### DIFF
--- a/src/core/connections.jl
+++ b/src/core/connections.jl
@@ -463,12 +463,6 @@ Add an array type parameter `name` with value `value` and `dims` dimensions to t
 """
 function set_external_array_param!(obj::AbstractCompositeComponentDef,
                                    name::Symbol, value::AbstractArray, dims)
-    numtype = Union{Missing, number_type(obj)}
-
-    if !(typeof(value) <: Array{numtype} || (value isa AbstractArray && eltype(value) <: numtype))
-        # Need to force a conversion (simple convert may alias in v0.6)
-        value = Array{numtype}(value)
-    end
     param = ArrayModelParameter(value, dims === nothing ? Vector{Symbol}() : dims)
     set_external_param!(obj, name, param)
 end
--- a/test/test_parametertypes.jl
+++ b/test/test_parametertypes.jl
@@ -51,6 +51,7 @@ eval(expr)  # Just a deprecation warning for v0.10, then will change to error in
     f = Parameter{Array{Float64, 2}}()
     g = Parameter{Int}(default=10.0)    # value should be Int despite Float64 default
     h = Parameter(default=10)           # should be "numtype", despite Int default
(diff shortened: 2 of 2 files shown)


jrising:
How does this work, if we do want to use autodifferentiation? Do things break if you set the model
`number_type` to `Number`?

The previous functionality (which admittedly was not considering all the cases) was that you could still set the parameters on a model as normal `Float64` arrays, and then they would be converted internally. That's useful for models that don't always want to be run in "autodiff" mode, and don't want to duplicate all of their parameters setting code to manually convert to Number. (At least, that's my recollection of the problems that emerged.)

ckingdon95:
@jrising if the autodiff functionality that you are describing was something that was previously working for arrays that have a time dimension, then it should still work. The arrays of values still get converted to the model's default number type in the outer function call (before the function `set_external_array_param!` that I've modified here is invoked). `set_external_array_param!` is called by `connect_param!` which is called by the original `set_param!`. It's in the `set_param!` function that the array of values provided by the user gets converted to the model's default number type, unless their is a type specified for that specific parameter (like `x::Int64=Parameter()`, in which case the values will stay as `Int`'s). What I've fixed here was just a bug where only for non-time arrays, the `set_external_array_param!` function was unnecessarily converting all values to the model's default type again (even though they had been converted in the outer `set_param!` function), but it was also doing so even in the case where the user had specified a parameter-specific type (this was the bug).

Not sure if that fully answers your question though, let me know if you still think there might be a problem for specifically using "Number" as the default type!

Another thing, if you want to test it out, please don't use this branch, but try it on the `param-type-0.9.6` branch. (This PR's branch is off of master, which has other breaking changes that you don't want yet, but the `param-type-0.9.6` branch will only include this fix for you to use)

ckingdon95:
here's an example with the number type set to "Number" that seems to still work:
```julia
using Mimi

@defcomp bar begin
    x::Int64 = Parameter(index=[regions])
    y = Parameter(index=[regions])
    z = Parameter(index=[time])
end

m = Model(Number)
set_dimension!(m, :time, 3)
set_dimension!(m, :regions, [:a, :b])
add_comp!(m, bar)
set_param!(m, :bar, :x, [1,2])
set_param!(m, :bar, :y, [1,2])    # even though y and z are provided as Int's, they will be converted to Number
set_param!(m, :bar, :z, [1,2,3])
run(m)
m[:bar,:x]    # returns an Array{Int64, 1}
m[:bar,:y]    # returns an Array{Number, 1}
m[:bar,:z]    # returns an Array{Number, 1}
```

jrising:
Thanks for testing this! Looks like it should work smoothly.

PR #665, opened 2020-02-28
Set param update

lrennels:
@ckingdon95 and @lrennels to fix the doctests, including adding `mim-next` branches to FUND and DICE2010, and any other tutorial changes we need.


Project.toml
```diff
@@ -37,7 +37,7 @@ VegaLite = "112f6efa-9a02-5b7d-90c0-432ed331239a"
 
 [compat]
 CSVFiles = "0.16, 1.0"
-Classes = "0.1"
+Classes = "1.1.0"
```
davidanthoff:
```suggestion
Classes = "1.2.0"
```

ckingdon95:
> Line 671: Is there a need to treat `count == 1` as and special case? Won't the loop will handle this? The only diff is `param_name` vs `nameof(param_ref)`, which I'm not sure are different.

For some reason, yes there are cases in some of the tests where these are different. I was just explaining to Lisa (but maybe should have said it in the general slack channel) that this fix only appears to work for flat models and not all composite models. It does work for some of the composites in the tests files though. I think it has to do with if the parameters have different names. This is definitely not a full/clean implementation though.

rjplevin:
I'm surprised it works since`connect_param!()` disconnects prior connections to the same external parameter name.

ckingdon95:
> I'm surprised it works since`connect_param!()` disconnects prior connections to the same external parameter name.

I think it only disconnects parameters for that component, not for all things bound to that external parameter. For example, the following works now (and works in the `mimi-next` branches on FUND and DICE2010):
```julia
@defcomp A begin
    p1 = Parameter()
end

@defcomp B begin
    p1 = Parameter()
end

m = Model()
set_dimension!(m, :time, 10)
add_comp!(m, A)
add_comp!(m, B)
set_param!(m, :p1, 5)

run(m)
m[:A, :p1] == m[:B, :p1] == 5
```

rjplevin:
Great -- maybe it's simpler than I expected!

lrennels:
Looks like we're failing a test in `test_timesteparrays.jl` where we fail to throw an exception.  Logging it here to look into.
```

@defcomp foo begin
    par = Parameter(index=[time])
    var = Variable(index=[time])
        if is_last(t)
            v.var[t] = 0
        else
            v.var[t] = p.par[t+1]   # This is where the error will be thrown, if connected to an internal variable that has not yet been computed.
        end
    end
end

years = 2000:2010

m = Model()
set_dimension!(m, :time, years)
add_comp!(m, foo, :first)
add_comp!(m, foo, :second)
connect_param!(m, :second => :par, :first => :var)
set_param!(m, :first, :par, 1:length(years))

@test_throws MissingException run(m)
```

PR #678, opened 2020-03-13
Add a test for external parameters

davidanthoff:
This is a small test that captures the problem that makes https://github.com/fund-model/MimiFUND.jl/pull/64 not pass right now.

First question: do we all agree that the behavior in this test is what we want? I think yes, right?

Next step, someone (other than me) pushes something to this branch here that makes the test pass :)

@rjplevin, @ckingdon95 and @lrennels .

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -17,6 +17,9 @@ Electron.prep_test_env()
     @info("test_main.jl")
     @time include("test_main.jl")
 
+    @info ("test_parameters.jl")
+    @time include("test_parameters.jl")
+
     @info("test_composite.jl")
     @time include("test_composite.jl")
 
--- a/test/test_parameters.jl
+++ b/test/test_parameters.jl
@@ -8,9 +8,12 @@ using Test
 end
 
 m = Model()
-
+set_dimension!(m, :time, 1:10)
 add_comp!(m, Foo)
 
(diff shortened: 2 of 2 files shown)


ckingdon95:
just making a note that I'll probably want to move this test into an existing test file (we have so many different test files!)

ckingdon95:
It's definitely a problem with parameters that have defaults. I added a line that shows that after you run the model (aka after it's been built), the parameter ends up in the external parameters list. I think it's actually a bit confusing what behavior we want in this case. When we add a component that has default parameters to a model, do we want those parameters to populate the model's external parameter dictionary at that moment?

rjplevin:
> When we add a component that has default parameters to a model, do we want those parameters to populate the model's external parameter dictionary at that moment?

Each `CompositeComponentDef` stores a vector of defaults. At build time (see `_set_defaults()` in build.jl) these are gathered into a `dict` that hashes on the component path and parameter name, and then applied to unconnected parameters.

I suspect that the leaf components may not be handled correctly in this processing.

ckingdon95:
Decided:
- you cannot call `update_param!` on something unless you have already used `set_param!` for that parameter
- so the current behavior of erroring is actually correct, and we need to change the test in FUND to use `set_param!`

PR #680, opened 2020-03-16
datum access method(s) for composites

ckingdon95:
WIP-- I think we want to implement methods for accessing parameters and variables that exist at the top level of a model with the following syntax:
```
m[:top_composite_name, :datum_name]
```
So far, I just added a test for this, and still need to implement it. We may want to remove the current methods for accessing values using a string for the component path. TBD.

### DIFF
--- a/test/test_composite.jl
+++ b/test/test_composite.jl
@@ -83,7 +83,6 @@ end
     fooA1 = Parameter(A.foo1)
     fooA2 = Parameter(A.foo2)
 
-    # TBD: component B isn't getting added to mi
     Component(B)
     foo3 = Parameter(B.foo3)
     foo4 = Parameter(B.foo4)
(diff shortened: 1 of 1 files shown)


rjplevin:
Note that `m[:top_composite_name][:datum_name]` should already work.

ckingdon95:
Both of the following work (using the example composite model from the test file I modified in this PR):
```julia
mi[:top][:A][:Comp2, :par_2_2]
mi["/top/A/Comp2", :par_2_2]
```
But none of the following work:
```julia
mi[:top][:par_2_2]
mi[:top, :par_2_2]
mi[:top][:fooA1]    # "fooA1" is what "par_2_2" was renamed as at the top level
mi[:top, :fooA1]
```
and I think we want something where we can access the parameters at the top level without needing to know the full path to the leaf component

Issue #657, opened 2020-02-25
Implement "joining" together of multiple parameters


rjplevin:
Separated out from #656.

Implement the `@defcomposite` option to join parameters in sub-components to a single name in a super-component. Proposed handling:

Joining parameters is permitted if:
-  the sub-component params are not otherwise connected
- the sub-component params have no default values, or all have the same default value

Once parameters are thus joined, it should not be permitted to set any of the individual sub-comp params directly, nor to connect any of them individually. That is, you can't do:
```
@defcomposite foo
    Component(x)
    Component(y)
    Component(z)
    joined_param = Parameter(x.p1, y.p1, z.p2)
    connect(y.p1, z.var1) # NOT ALLOWED
end
```
This also suggests means disallowing set_param!(composite, ...) to be called on parameters referenced by joined parameters. If we only allow set_param!(m, ...) this is fine, but we need to remove deprecated methods, e.g., that operate on component paths or component path strings like "/foo/bar".

_Note: A version of the following appears as comments appear in types/defs.jl, lines 116:141._

Rather than store a Vector of `ParameterDefReference`, I think instances of this class should store multiple references, all of which share one (local) name and at most one default
 value. The refactored classes might look like this:
```
# This provides a common point (AbstractDatumReference) for referring to both 
# param and var references.
@class DatumReference <: NamedObj

struct UnnamedReference
   root::AbstractComponentDef
   comp_path::ComponentPath
end

@class ParameterDefReference <: DatumReference begin
   # N.B. name::Symbol is inherited from NamedObj
   default::Any    # allows defaults set in composites
   refs::Vector{UnnammedReference} # one or more parameter references
end

@class VariableDefReference  <: DatumReference begin
   # N.B. name::Symbol is inherited from NamedObj
   unnamed_ref::UnnamedReference  # a single variable reference
end
```

The main job is reviewing, and, where needed, revising any code that touches these classes. The other important piece is reviewing cases of parameter setting or connecting to ensure none of the restrictions above are violated.

ckingdon95:
notes from 3/19:
- I'll implement this as described by Rich above
- possibly rename second two types to match how they are viewed from the outside, something like "CompositeParameterDef". These are actually definitions of parameters and variables in the composites.
- CompositeParameterDef and LeafParameterDef etc. would inherit DatumDef, all should have optional keyword arguments for defaults, units, and descriptions. One can override at the composite level. (if they are different at the leaf level and aren't overridden, then error)

Issue #524, opened 2019-06-11
Changes to set_param!


davidanthoff:
The core functions are:
```julia
set_param!(m, :compA, :p1, :ext_p2_name, 34.)
update_param!(m, :ext_p2_name, 45.)
```

And the shortcuts are:
```julia
set_param!(m, :p1, 23.)
```
This function would look at all the components in the model, find all the unbound parameters named `p1`, create an external parameter called `p1`, set its value to `23`, and create connections from all the external unbound parameters that are named `p1` to the external parameter `p1`. If there is already a `p1` in the external parameter list, it errors.

```julia
set_param!(m, :compA, :p1, 34.)
```
This function creates a new external parameter called `p1` (if that already exists, it errors), sets its value to `34.`, and then connects `compA/p1` to this external parameter `p1`.

davidanthoff:
Maybe also need:
```julia
set_param!(m, [:compA, :compC, :compD], :p1, 23)
set_param!(m, [:compA=>:p1, :compC=>:p3], :p_ext_3, 34.)
connect_param!(m, :compA, :p1, :p_ext_3)
```

davidanthoff:
`set_param!` is not a good name for all of this. So we'll come up with new better names for the new functions we creating. We'll leave `set_param!` around as is with the old behavior, except, that we make 
```julia
set_param!(m, :compA, :p1, 34.)
set_param!(m, :compB, :p1, 12.)
```
error.

Rollout is: we add the new fancy functions with better names first, tag a release. Then we migrate models over to that. Then we start to throw errors in a future Mimi release for the buggy `set_param!` corner case.

rjplevin:
@davidanthoff, The current design defines `external_params` in the `CompositeComponentDef`, so these can exist at all levels, including `ModelDef`.

This is necessary since a component might have external params and it can't be aware of external params of the same name defined elsewhere in the model. To connect these knowledgeably, I suggest we provide a function that lists all the external params of a given name, ideally with a documentation string, so the modeler can learn whether all the params named `:foo` serve the same purpose prior to connecting/setting them all.

Should this method apply only to unbound params? What if you want to join together parameters `:foo` that has the same semantics across multiple components? They might already be bound, but connecting them to a single external param at a higher composite level makes sense.

[...]

PR #687, opened 2020-03-26
Implement "joining" multiple parameters in @defcomposite

ckingdon95:
#657 
and I've also implemented the various `set_param!` methods described in #524


src/core/build.jl
```diff
@@ -253,11 +227,9 @@ function _build(md::ModelDef)
 end
 
 function build(m::Model)
-    # import any unconnected params into ModelDef
-    import_params!(m.md)
 
     # apply defaults to unset parameters
-    _set_defaults!(m.md)
+    _set_defaults!(m.md)    # should this actually happen on the deepcopy below isntead?
```
ckingdon95, line 232: need to discuss this or remove comment

src/core/connections.jl
```diff
@@ -278,94 +278,152 @@ function split_datum_path(obj::AbstractCompositeComponentDef, s::AbstractString)
...
-function connect_param!(obj::AbstractCompositeComponentDef, dst::AbstractString, src::AbstractString,
-                        backup::Union{Nothing, Array}=nothing; ignoreunits::Bool=false, offset::Int=0)
-    dst_path, dst_name = split_datum_path(obj, dst)
-    src_path, src_name = split_datum_path(obj, src)
+function _find_paths_and_names(obj::AbstractComponentDef, datum_name::Symbol)
+
+    # Base case-- leaf component
+    if obj isa ComponentDef
+        return ([nothing], [datum_name])
+    end
```
ckingdon95, line 292: note to self: might clean up this base case to actually return the comp name instead of nothing, after the comp_path bug is fixed separately

src/core/build.jl
```diff
@@ -92,6 +92,100 @@ function _instantiate_vars(md::ModelDef)
...
+
+Recurses through sub components and finds the full path(s) to desired datum, and their
+names at the leaf level. Returns a tuple (paths::Vector{ComponentPath}, datum_names::Vector{Symbol})
+"""
+function _find_paths_and_names(obj::AbstractComponentDef, datum_name::Symbol)
+
+    # Base case-- leaf component
+    if obj isa ComponentDef
+        return ([nothing], [datum_name])
+    end
```
ckingdon95, line 106: note to self: might clean up this base case to actually return the comp name instead of nothing, after the comp_path bug is fixed separately

Issue #690, opened 2020-04-02
improve Int time indexing warning message


ckingdon95:
@lrennels could you look at this? I think you implemented the warning message for deprecating Int indexing. It seems like the message prints out the whole component instance type or something, which makes for a pretty long message. It looks like something like this in PAGE (looks worse in the terminal with word wrap):
```
[...]

PR #694, opened 2020-04-05
Improve int indexing depwarn msg

lrennels:
This PR addresses #690, and improves the error message for the deprecation warning associated with integer indexing. Since the Stacktrace will change in each call, the improvement for now prints out the `stacktrace()` stack frame line by line instead of as a text block, ie. the following.

The component printout could still be quite long, however, but at least printed nicely.

```
┌ Warning: Indexing with getindex into a TimestepArray with Integer(s) is deprecated, please index with a TimestepIndex(index::Int) instead ie. instead of t[2] use t[TimestepIndex(2)]
│  
│ _get_stacktrace_string() at time_arrays.jl:62
│ _throw_int_getindex_depwarning at time_arrays.jl:71 [inlined]
│ getindex(::TimestepArray{FixedTimestep{2000,1,LAST} where LAST,Int64,1,1}, ::Int64) at time_arrays.jl:209
│ top-level scope at REPL[68]:1
│ eval(::Module, ::Any) at boot.jl:331
│ eval_user_input(::Any, ::REPL.REPLBackend) at REPL.jl:86
│ macro expansion at REPL.jl:118 [inlined]
│ (::REPL.var"#26#27"{REPL.REPLBackend})() at task.jl:358
│   caller = top-level scope at REPL[68]:1
└ @ Core REPL[68]:1
```

### DIFF
--- a/src/core/time_arrays.jl
+++ b/src/core/time_arrays.jl
@@ -56,16 +56,29 @@ function _single_index_check(data, idxs)
 	end
 end
 
+# Helper to print stacktrace
+function _get_stacktrace_string()
+	s = ""
+	for line in stacktrace()
+		s = string(s, line, "\n")
+	end
+	return s
+end
+
 # Helper functionfor getindex; throws an error if one indexes into a TimestepArray with an integer
 function _throw_int_getindex_depwarning()
-	msg = "Indexing with getindex into a TimestepArray with Integer(s) is deprecated, please index with a TimestepIndex(index::Int) instead ie. instead of t[2] use t[TimestepIndex(2)]"
-	Base.depwarn("$msg, $(stacktrace())", :getindex)
+	msg = "Indexing with getindex into a TimestepArray with Integer(s) is deprecated, please index with a TimestepIndex(index::Int) instead ie. instead of t[2] use t[TimestepIndex(2)]\n"
+	st = _get_stacktrace_string()
+	full_msg = string(msg, " \n", st)
+	Base.depwarn(full_msg, :getindex)
 end
 
 # Helper function for setindex; throws an deprecation warning if one indexes into a TimestepArray with an integer
 function _throw_int_setindex_depwarning()
 	msg = "Indexing with setindex into a TimestepArray with Integer(s) is deprecated, please index with a TimestepIndex(index::Int) instead ie. instead of t[2] use t[TimestepIndec(2)]"
-	Base.depwarn("$msg, $(stacktrace())", :setindex)
+	st = _get_stacktrace_string()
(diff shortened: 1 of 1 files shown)


lrennels:
@davidanthoff @ckingdon95 should we leave the whole stacktrace, or nab just a part of it?

ckingdon95:
Although I think the whole component instance type might still get printed out on the "macro expansion" line. Have you tried it with the mimi-next branch of PAGE? does this make it any less long?

ckingdon95:
actually maybe we can implement something even sneakier: if the error message is coming from inside of a run_timestep function (which 99% of these should be) then could your `_get_stacktrace_string` function check for that? aka I think there will always be a line that starts with 
```
macro expansion at EquityWeighting.jl:116 [inlined]
```
followed by a line like: 
```
run_timestep_MimiPAGE2009_EquityWeighting(::Mimi.ComponentInstanceParameters{NamedTuple . . .
# this is the really long one
```
So maybe we can drop any lines that start with "run_timestep" so that it would just return the previous line, the one that actually says the line number in the component code?

lrennels:
@ckingdon95 yea that's exactly what I was thinking of doing, return the **first** line that starts with `macro expansion` I just wasn't sure if that seemed too fragile

ckingdon95:
I think that makes sense to only return that line. But then if there aren't any lines that start with "macro expansion..." return the whole thing

lrennels:
Ok deal we'll have it return the first macro expansion line, and if there aren't any it'll return the whole trace

ckingdon95:
@lrennels there are definitely tests in the test files where this arises outside of a "run_timestep" function, but I meant that in terms of actual users I think this warning will 99% be coming from "run_timestep" functions.

src/core/time_arrays.jl
```diff
@@ -56,11 +56,17 @@ function _single_index_check(data, idxs)
...
 
-# Helper to print stacktrace
+# Helper to print stacktrace for the integer indexing deprecatino warning
 function _get_stacktrace_string()
 	s = ""
 	for line in stacktrace()
-		s = string(s, line, "\n")
+		line_string = string(line)
+		if startswith(line_string, "run_timestep") # shorten the reporting on calls to run_timestep
+			l = "run_timestep function call"
```
ckingdon95, line 65: My preference would be to just `return s` here. If it hits a "run_timestep" line, then the previous line would be the "macro expansion" line which has the actual component line number info of where the error is happening

lrennels: @ckingdon95 sounds good to me, done!

Issue #689, opened 2020-04-01
bug in `_fix_comp_path` for internal_param_conns


ckingdon95:
There seems to be a bug (on master) in the way that the component paths in subcomponents' internal_param_conns get updated, but the bug only arises after three levels of composites. Here is an example:
```julia
using Mimi

@defcomp A begin
    v1 = Variable()
end

@defcomp B begin
    p1 = Parameter()
end

@defcomposite C begin
    Component(A)
    Component(B)
    connect(B.p1, A.v1)   # make an internal connection in comp C
end

@defcomposite D begin
    Component(C)
end

@defcomposite E begin
    Component(D)
end
```
This displays the correct src_comp_path and dst_comp_path:
```
julia> C.internal_param_conns
1-element Array{Mimi.InternalParameterConnection,1}:
 Mimi.InternalParameterConnection
  src_comp_path: ComponentPath(:C, :A)
  src_var_name: :v1
  dst_comp_path: ComponentPath(:C, :B)
  dst_par_name: :p1
  ignoreunits: false
  backup: nothing
  offset: 0
```
This one is also right:
```
julia> D[:C].internal_param_conns
1-element Array{Mimi.InternalParameterConnection,1}:
 Mimi.InternalParameterConnection
  src_comp_path: ComponentPath(:D, :C, :A)
  src_var_name: :v1
  dst_comp_path: ComponentPath(:D, :C, :B)
  dst_par_name: :p1
  ignoreunits: false
  backup: nothing
  offset: 0
```
But this one is messed up:
```

julia> E[:D][:C].internal_param_conns
1-element Array{Mimi.InternalParameterConnection,1}:
 Mimi.InternalParameterConnection
  src_comp_path: ComponentPath(:E, :D, :C, :C, :A)
  src_var_name: :v1
  dst_comp_path: ComponentPath(:E, :D, :C, :C, :B)
  dst_par_name: :p1
  ignoreunits: false
  backup: nothing
  offset: 0
```

rjplevin:
Let me know if you guys want me to address this.

ckingdon95:
Actually, it would be great if this is something that you could fix! Presumably it's a quick, somewhat isolated fix (but who knows...) would you be able to look at it soon?

davidanthoff:
Yes, if this is an easy fix it would be great if you (@rjplevin) could do it!

rjplevin:
And if it's not an easy fix? ;~)

[...]

PR #697, opened 2020-04-09
fix `_fix_comp_path!` for internal connections

ckingdon95:
#689 
The problem was that `tail(comp_path)` returns not just the last element, it returns everything but the head. But what we want here is just the last element.

### DIFF
--- a/src/core/paths.jl
+++ b/src/core/paths.jl
@@ -47,8 +47,8 @@ function _fix_comp_path!(child::AbstractComponentDef, parent::AbstractCompositeC
         # Fix internal param conns
         conns = child.internal_param_conns
         for (i, conn) in enumerate(conns)
-            src_path = ComponentPath(child_path, tail(conn.src_comp_path))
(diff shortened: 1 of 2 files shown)


src/core/paths.jl
```diff
@@ -47,8 +47,8 @@ function _fix_comp_path!(child::AbstractComponentDef, parent::AbstractCompositeC
         # Fix internal param conns
         conns = child.internal_param_conns
         for (i, conn) in enumerate(conns)
-            src_path = ComponentPath(child_path, tail(conn.src_comp_path))
-            dst_path = ComponentPath(child_path, tail(conn.dst_comp_path))
+            src_path = ComponentPath(child_path, conn.src_comp_path.names[end])
+            dst_path = ComponentPath(child_path, conn.dst_comp_path.names[end])
```
ckingdon95, line 51: I suppose I also could have done this instead
```
 src_path = ComponentPath(head(child_path), conn.src_comp_path)
```

But since this works I guess it doesn't matter....

Issue #686, opened 2020-03-24
using renamed components in @defcomposite


ckingdon95:
The following renaming of leaf component A works:
```julia
using Mimi

@defcomp A begin
    p1 = Parameter()
end

@defcomposite B begin
    foo = Component(A)
end
```
But if you try to use the renamed component in a parameter definition, it errors:
```julia
@defcomposite C begin
    foo = Component(A)
    rename_p1 = Parameter(foo.p1)
end
```
```
ERROR: LoadError: UndefVarError: foo not defined
Stacktrace:
 [1] top-level scope at C:\Users\kingdon\.julia\dev\Mimi\src\core\defcomposite.jl:183
 [2] include_string(::Module, ::String, ::String) at .\loading.jl:1075
 [3] #invokelatest#1 at .\essentials.jl:709 [inlined]
 [4] invokelatest at .\essentials.jl:708 [inlined]
 [5] (::Main._vscodeserver.var"#9#12"{String,Int64,Int64,String})() at c:\Users\kingdon\.vscode\extensions\julialang.language-julia-0.14.18\scripts\terminalserver\terminalserver.jl:162
 [6] withpath(::Main._vscodeserver.var"#9#12"{String,Int64,Int64,String}, ::String) at c:\Users\kingdon\.vscode\extensions\julialang.language-julia-0.14.18\scripts\terminalserver\repl.jl:62
 [7] (::Main._vscodeserver.var"#8#11"{String,Int64,Int64,String})() at c:\Users\kingdon\.vscode\extensions\julialang.language-julia-0.14.18\scripts\terminalserver\terminalserver.jl:161
 [8] hideprompt(::Main._vscodeserver.var"#8#11"{String,Int64,Int64,String}) at c:\Users\kingdon\.vscode\extensions\julialang.language-julia-0.14.18\scripts\terminalserver\repl.jl:28
 [9] macro expansion at c:\Users\kingdon\.vscode\extensions\julialang.language-julia-0.14.18\scripts\terminalserver\terminalserver.jl:152 [inlined]   
 [10] (::Main._vscodeserver.var"#7#10")() at .\task.jl:333
in expression starting at c:\Users\kingdon\.julia\dev\Mimi\example.jl:7
```
This error already exists on master currently, so I think I might address it later instead of during my current re-write of composite parameter definitions. (i.e. I'll get the parameter stuff to work first without being able to rename components)

lrennels:
#698

PR #698, opened 2020-04-10
using renamed components in #defcomposite

lrennels:
#686


ckingdon95:
I think this error will also need to be addressed not just for `Parameter` statements but also `Variable` statements and `connect` statements. Here's an expanded example to hit all three:
```julia
@defcomp A begin
    p1 = Parameter()
    p2 = Parameter()

    v1 = Variable()
end

@defcomp B begin
    v2 = Variable()
end

@defcomposite C begin

    foo = Component(A)
    bar = Component(B) 

    rename_p1 = Parameter(foo.p1) 

    connect(foo.p2, bar.v2)

    rename_v1 = Variable(foo.v1)
end
```

lrennels:
@ckingdon95 the macro yields the following with the `@defcomposite C` call (in a let block but I removed that for now)

```
cc_id = Mimi.ComponentId(Main, :C)
obj = Mimi.CompositeComponentDef(cc_id)

#= /Users/lisarennels/.julia/dev/Mimi/src/core/defcomposite.jl:183 =#
global C = obj

#= /Users/lisarennels/.julia/dev/Mimi/src/core/defcomposite.jl:184 =#
Mimi.add_comp!(obj, A, :foo)
Mimi.add_comp!(obj, B, :bar)
Mimi.import_param!(obj, :rename_p1, obj[:foo] => :p1; )
Mimi.connect_param!(obj, :foo, :p2, :bar, :v2; )
Mimi._import_var!(obj, :rename_v1, ComponentPath(:foo,), :v1)

#= /Users/lisarennels/.julia/dev/Mimi/src/core/defcomposite.jl:185 =#
Mimi.import_params!(obj)

#= /Users/lisarennels/.julia/dev/Mimi/src/core/defcomposite.jl:186 =#
C
```

lrennels:
Note from conversation: This macro expand looks good, next step is to move the wip file to testing, perhaps test for keys like: 

```
@test (:foo, :bar, :rename_p1, :rename_v1) in keys(C.namespace)
```

Issue #677, opened 2020-03-12
fix getindex methods for composites


ckingdon95:
We need to be able to access the parameters and variables that are explicitly available at the composite level, probably via the following:
```
m[:top_composite_name, :exposed_variable_name]
```
I think we definitely need that method working for v1.0.0, which currently errors.

We had also discussed the ability to use dot-syntax to go down the tree to access values that maybe aren't bubbled up? something like:
```
m.top_composite.leaf_component.variable_name
```
But maybe that's not necessary for v1.0.0

ckingdon95:
@lrennels I found this original issue for datum access methods, so I closed the other one

lrennels:
#704

PR #704, opened 2020-04-16
Datum access

lrennels:
#677 

Currently we have implemented the methods and constructors needed to access parameters and variables in the top level composite components of a model with a call like

```
m[:top, :par_name
```
next steps are to add more comprehensive testing and think about corner cases like renaming.  also, do we want to be able to use this syntax for lower levels i.e. if `:B` is a child composite component of `:top` but not in the `comps_dict` of `m.mi`, should we be able to do 
```
m[:B, :par name]
```

### DIFF
--- a/src/core/build.jl
+++ b/src/core/build.jl
@@ -241,9 +241,10 @@ function _instantiate_params(comp_def::ComponentDef, par_dict::Dict{Tuple{Compon
     return ComponentInstanceParameters(names, types, vals, paths)
 end
 
-function _instantiate_params(comp_def::AbstractCompositeComponentDef, par_dict::Dict{Tuple{ComponentPath, Symbol}, Any})
-    _combine_exported_pars(comp_def)
-end
+# LFR: commented out because potentially doing nothing ... never called
+# function _instantiate_params(comp_def::AbstractCompositeComponentDef, par_dict::Dict{Tuple{ComponentPath, Symbol}, Any})
+#     _combine_exported_pars(comp_def)
+# end
 
 # Return a built leaf or composite LeafComponentInstance
 function _build(comp_def::ComponentDef,
@@ -269,6 +270,11 @@ function _build(comp_def::AbstractCompositeComponentDef,
     # @info "  par_dict $(par_dict)"
 
     comps = [_build(cd, var_dict, par_dict, time_bounds) for cd in compdefs(comp_def)]
+
+    # LFR TODO
+    # vars_dict = get_vars_dict(comp_def)
+    # pars_dict = get_pars_dict(comp_def)
+    
     return CompositeComponentInstance(comps, comp_def, time_bounds)
 end
 
--- a/src/core/build.jl
+++ b/src/core/build.jl
@@ -271,11 +271,26 @@ function _build(comp_def::AbstractCompositeComponentDef,
 
     comps = [_build(cd, var_dict, par_dict, time_bounds) for cd in compdefs(comp_def)]
 
-    # LFR TODO
-    variables = NamedTuple(Tuple(:var1, :var2), Tuple(1.0, 2.0))
-    parameters = NamedTuple(Tuple(:par1, :par2), Tuple(1.0, 2.0))
+    variables = _get_variables(comp_def, var_dict)
+    parameters = _get_parameters(comp_def, par_dict)
 
     return CompositeComponentInstance(comps, comp_def, time_bounds, variables, parameters)
+end
+
+# helper functions for to create the variables and parameters NamedTuples for a 
+# CompositeComponentInstance
+function _get_variables(comp_def::AbstractCompositeComponentDef, var_dict::Dict{ComponentPath, Any})
+
+    # TODO (dummy below)
+    NT = NamedTuple(Tuple(:var1, :var2), Tuple(1.0, 2.0))
+    return NT
+end
+
+function _get_parameters(comp_def::AbstractCompositeComponentDef, par_dict::Dict{Tuple{ComponentPath, Symbol}, Any})
+
+    # TODO (dummy below)
+    NT = NamedTuple(Tuple(:par1, :par2), Tuple(1.0, 2.0))
+    return NT
 end
 
 """
--- a/src/core/build.jl
+++ b/src/core/build.jl
@@ -241,7 +241,7 @@ function _instantiate_params(comp_def::ComponentDef, par_dict::Dict{Tuple{Compon
     return ComponentInstanceParameters(names, types, vals, paths)
 end
 
-# LFR: commented out because potentially doing nothing ... never called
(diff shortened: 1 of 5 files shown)


src/core/build.jl
```diff
@@ -269,7 +270,27 @@ function _build(comp_def::AbstractCompositeComponentDef,
...
+
+    return CompositeComponentInstance(comps, comp_def, time_bounds, variables, parameters)
+end
+
+# helper functions for to create the variables and parameters NamedTuples for a 
+# CompositeComponentInstance
+function _get_variables(comp_def::AbstractCompositeComponentDef, var_dict::Dict{ComponentPath, Any})
+
+    # TODO (dummy below)
+    NT = NamedTuple(Tuple(:var1, :var2), Tuple(1.0, 2.0))
```
lrennels, line 285: @ckingdon95 advice here would be great!

src/core/build.jl
```diff
@@ -269,7 +270,27 @@ function _build(comp_def::AbstractCompositeComponentDef,
...
+    return NT
+end
+
+function _get_parameters(comp_def::AbstractCompositeComponentDef, par_dict::Dict{Tuple{ComponentPath, Symbol}, Any})
+
+    # TODO (dummy below)
+    NT = NamedTuple(Tuple(:par1, :par2), Tuple(1.0, 2.0))
+    return NT
```
lrennels, line 293: and here!

ckingdon95: and the only difference for the parameters is that CompositeParameterDef's can have multiple refs (they have a `refs` field instead of a `ref` field). As we discussed, all of them will have the same values at run time, so you can just use the first one `par_def.refs[1]`

src/core/build.jl
ckingdon95: so all this needs to do is loop through the comp_def's namespace (which is a dictionary) and keep everything that is a CompositeVariableDef. Then the symbols in the NameTuple constructor will just be the keys from the namespace, and the values in the second half of the constructor will be the vardef.ref (which gives an UnnamedReference object) from each vardef

```
# pseudo code

var_defs = filter(isa(CompositeVariableDef), com_pdef.namespace)  # this probably doesn't quite work for filtering a dictionary on it's values
symbols = [k for (k, v) in var_defs]
vals = [v.ref for (k, v) in var_defs]
NT = NamedTuple(symbols, vals)

File: src/core/instances.jl [hunk 3]
```diff
@@ -168,10 +168,37 @@ function Base.getindex(obj::AbstractCompositeComponentInstance, comp_name::Symbo
...
+        else
+            error("$datum_name is not a parameter or a variable in component $(ci.comp_path).")
+        end
+    end
+    
+    ref = getproperty(which, datum_name)
+    leaf_comp_name = ref.comp_name
+    leaf_datum_name = ref.datum_name
+    
+    # LFR TODO figure out which path to go down to find the leaf
```
lrennels, line 191: @ckingdon95 as we discussed, we need to just get the next component and datum name to recursively call on, not the final ones represented in the `UnnamedReference`.  I think we could sort of BFS/DFS down to find the leaf and then return it, but wanted to check in to see if you had ideas about where the info would be saved?

src/core/instances.jl
```diff
@@ -168,10 +168,33 @@ function Base.getindex(obj::AbstractCompositeComponentInstance, comp_name::Symbo
...
+        if datum_name in keys(pars) # could merge with below if made names(NamedTuple) = keys(NamedTuple)
+            which = pars
+        else
+            error("$datum_name is not a parameter or a variable in component $(ci.comp_path).")
+        end
+    end
+    
+    ref = getproperty(which, datum_name)
+    
+    # LFR TODO get the comp_name and datum_name to use ...
```
lrennels, line 189: @ckingdon95 this is a little bit tricky, because sometimes `UnnamedReference` skips a layer, so the `ref.comp_name` is not held in the `ci.comps_dict` ie. in the test we have currently we have a `CompositeComponentInstance` where parameters are fine:
```
julia> ci.parameters 
(fooA1 = Mimi.UnnamedReference(:A, :foo1), 
fooA2 = Mimi.UnnamedReference(:A, :foo2), 
foo3 = Mimi.UnnamedReference(:B, :foo3),
foo4 = Mimi.UnnamedReference(:B, :foo4), 
par_1_1 = Mimi.UnnamedReference(:A, :par_1_1))
```
but the variable just goes all the way to `:Comp3`
```
julia> ci.variables
(var_3_1 = Mimi.UnnamedReference(:Comp3, :var_3_1),)
```
and thus calling `_get_datum(ref.comp_name, ref.datum_name)` won't work...

src/core/instances.jl
```diff
@@ -168,10 +168,33 @@ function Base.getindex(obj::AbstractCompositeComponentInstance, comp_name::Symbo
...
+    return _get_datum(ci.comps_dict[comp_name, datum_name])
```
lrennels, line 190: I hoped this could just be `_get_datum(ci.comps_dict[ref.comp_name], ref.datum_name)` but see comment above

src/core/build.jl
```diff
@@ -241,9 +241,10 @@ function _instantiate_params(comp_def::ComponentDef, par_dict::Dict{Tuple{Compon
     return ComponentInstanceParameters(names, types, vals, paths)
 end
 
-function _instantiate_params(comp_def::AbstractCompositeComponentDef, par_dict::Dict{Tuple{ComponentPath, Symbol}, Any})
-    _combine_exported_pars(comp_def)
-end
+# commented out because potentially doing nothing ... the fcn is never called
+# function _instantiate_params(comp_def::AbstractCompositeComponentDef, par_dict::Dict{Tuple{ComponentPath, Symbol}, Any})
+#     _combine_exported_pars(comp_def)
+# end
```
ckingdon95, line 247: Can you delete this function and also the `_combine_exported_pars` function too?

test/test_composite.jl
```diff
@@ -138,14 +138,30 @@ run(m)
...
 @test mi["/top/B/Comp4", :par_4_1] == collect(6.0:6:96.0)
 
-#
+@test m[:top, :fooA1] == 1
+@test m[:top, :var_3_1] == collect(6.0:6:96.0)
+
+# vars
+
 # Test joining external params.
 #
```
ckingdon95: Can you clean these comment lines? I think the # got deleted for the next block, and maybe you added `# vars` not needed anymore?

lrennels:
Hmm so would the idea of not naming them `variable` and `parameter` be that they are difference since they are immutable at this level and just references to lower components as opposed to unique ones?

ckingdon95:
yeah mostly the second point-- just that they don't actually store parameter/variable values, but references to them. minor point, but I feel like David would have a clear opinion on it so we could just revisit it later!

Issue #530, opened 2019-07-14
Rename `marginal` attribute of MarginalModel


jrising:
The attributes of the `MarginalModel` object are confusingly (and incorrectly) named. The `marginal` attribute does not contain itself some kind of marginal model object, or anything like it. Rather, it contains a perturbed model. I propose that the `marginal` attribute be renamed `perturbed`.

davidanthoff:
Another option might be to rename the fields to `model1` and `model2`.

ckingdon95:
or `base` and `modified`.

davidanthoff:
@jrising do you have a preference for one of these options? We are all indifferent here, but happy to change it.

jrising:
I think that `base` and `modified` are sensible, if we don't want to assume that a `MarginalModel` was created only by the SCC process. That would be my preference.

lrennels:
This would be breaking, so we we should make a call here.

davidanthoff:
Indeed. I thought we had made a decision at some point? I have to admit that I don't care _that_ much. But, whatever we decide now will stick :)

ckingdon95:
Yeah we decided to go with "modified". I will do this, it will be quick, I was just going to wait until Rich's stuff is merged to make it even easier!

jrising:
Thanks for doing this! I don't mind the breaking change, but it might be nice to do a "gradual deprecation", where the `marginal` attribute becomes an alias for `modified` that also prints out a warning telling people to switch over before the release after this one.

ckingdon95:
Sounds good, we will do the deprecated version as well

arnavgautam:
I can work on this

PR #706, opened 2020-05-01
Rename the 'marginal' attribute of MarginalModel to 'modified'

arnavgautam:
Renames the attribute, and adds an alias for the old 'marginal' name which includes a deprecation warning
See  https://github.com/mimiframework/Mimi.jl/issues/530


arnavgautam:
Ah yeah I had done manual testing by keeping one of the old lines and running the test suite. The line showed up as a properly formatted warning, but the tests still passed. If you want, I can put that back in, but I figured users would be confused if it showed up for them.

ckingdon95:
The value of keeping it in would be to make sure that the new function you added gets tested properly, but I'm surprised to see that the coverage tests that ran say that 100% of the diff is hit... I would've thought that it would miss those lines... Maybe it's possible that the model tests (i.e. DICE and FUND) that get run by the test suite are hitting those lines of the function... Either way I think it would make sense to just leave one in the test file, and add a comment that says you are testing the soon-to-be-deprecated behavior

lrennels:
Yea I agree, we've done that with some of the other deprecated code, just do leave a note and it'll be easy for us to find it later!

arnavgautam:
Updated it with the test, sorry for the git weirdness but I guess the VSCode interface messed me up

ckingdon95:
@lrennels do you know why some of your most recent commits are showing up here in the diffs of the "Files changed" even though they've already been commited to master?

lrennels:
@ckingdon95 Oh that's strange no I don't know, I was thinking maybe if I approved it would re-compare to master but it didn't. Odd ...

lrennels:
@ckingdon95 I just merged a little toml update so I'll update this branch and see if that helps

ckingdon95:
awesome thanks Lisa!

PR #711, opened 2020-06-02
Explorer for Composite Components

lrennels:
Make small changes to allow a call to `explore(m)` on a model that contains composite components.  For now this will only display the namespaces of the top level components, and will not dive into the sub-components of a composite component.

#711

### DIFF
--- a/src/core/model.jl
+++ b/src/core/model.jl
@@ -205,7 +205,7 @@ Return an iterator on the components in a model's model instance.
 
 Return a DatumDef for `item` in the given component `comp_def`.
 """
-function datumdef(comp_def::ComponentDef, item::Symbol)
+function datumdef(comp_def::AbstractComponentDef, item::Symbol)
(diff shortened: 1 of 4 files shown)


src/core/model.jl
```diff
@@ -205,7 +205,7 @@ Return an iterator on the components in a model's model instance.
 
 Return a DatumDef for `item` in the given component `comp_def`.
 """
-function datumdef(comp_def::ComponentDef, item::Symbol)
+function datumdef(comp_def::AbstractComponentDef, item::Symbol)
```
lrennels, line 208: @ckingdon95 do you see a downside to doing it this way (generalizing to one signature for all `AbstractComponentDef` types  instead of making a separate method for a `CompositeComponentDef`?  All tests pass, but you have a better sense of the theory behind this so I want to make sure I'm not breaking something.

Issue #585, opened 2019-11-01
turn deprecation warnings into errors


ckingdon95:
Steps:
1. Release v0.10.0 with the following depreciation warnings (using either `@deprecate` or `Base.depwarn`
2. Release v1.0.0 where these now error, mostly with a `MethodException` since methods will be removed. This will require cleaning out the deprecated code, any helper functions for deprecation warnings, and any tests of the now removed code.

Deprecations include the following (also search for `Base.depwarn` and `@deprecate`)

- Integer indexing into a `TimestepArray` at the time index position with `getindex` or `setindex!`
- using the functions `is_time` or `is_timestep` 
- indexing into a simulation instance with square brackets 
- changes that need to be made to `set_param!`
- type parameterization in @defcomp #638 
- renaming `marginal` to `modified` in `MarginalModel`'s
- change `replace_comp!` to `replace!` w/ julia argument convention (use pair)

lrennels:
I think it's probably a good idea to add deprecation warnings to anything where it's a somewhat one to one change with functions, and then the rest of the things could be in a nice comprehensive list or something that we can post on the website/docs.

lrennels:
In terms of indexing into a simulation index with square brackets, I have completely removed that `getindex` function and replaced it with a `getdataframe`, so we'd have to add the `getindex` back in if we wanted to deprecate it.  I have a feeling no one is using that so it might not be necessary but either way would be fine.

rjplevin:
If nobody is using it, the deprecation does no harm. If they are using it and you don't deprecate, it causes grief. Any reason to not deprecate? Seems like the better choice to me!

lrennels:
Agreed, no reason not to deprecate except laziness, I'll go grab that old PR and add add back in a deprecated `getindex` function!

lrennels:
#743

PR #713, opened 2020-06-13
Remove deprecations

lrennels:
This PR addresses #585 and transitions from v0.10.0 to v1.0.0 by removing deprecation warning code so that deprecated syntax will error. Note that a small drop in coverage is inevitable since we are leaving in a few functions but no longer supporting, and thus no longer testing, them.


src/core/defcomp.jl
```diff
@@ -195,10 +195,7 @@ macro defcomp(comp_name, ex)
             continue
         end
 
-        if @capture(elt, name_::datum_type_ = elt_type_(args__))
-            msg = "The following syntax has been deprecated in @defcomp: \"$name::$datum_type = $elt_type(...)\". Use curly bracket syntax instead: \"$name = $elt_type{$datum_type}(...)\""
-            Base.depwarn("$msg\n$(reduce((x,y) -> "$x\n$y", stacktrace()))", :defcomp)   
-        elseif ! @capture(elt, name_ = (elt_type_{datum_type_}(args__) | elt_type_(args__)))
+        if ! @capture(elt, name_ = (elt_type_{datum_type_}(args__) | elt_type_(args__)))
```
lrennels, line 198: @ckingdon95 does this look right I'm bad at macros

src/core/time.jl
```diff
@@ -20,16 +20,6 @@ function gettime(ts::VariableTimestep)
...
-"""
-	is_time(ts::AbstractTimestep, t::Int)
-
-Return true or false, true if the current time (year) for `ts` is `t`
-"""
-function is_time(ts::AbstractTimestep, t::Int) 
-	Base.depwarn("`is_time(ts, t)` is deprecated. Use comparison operators with TimestepValue objects instead: `ts == TimestepValue(t)` \n$(stacktrace())", :is_time)
-	return gettime(ts) == t
-end
-
```
ckingdon95, line 32: same question for the rest of these as well: do we just want to delete the old functionality entirely, or should we just turn it into helpful error messages?

src/core/defcomp.jl
lrennels: Yea I was struggling with that for most of the deprecations to, I’m not sure where we would use a special error message and where we just delete the method so it ends up as a method error (like with the timestep indexing with integers).

src/core/time.jl
lrennels: Agreed I’m not entirely sure what the intention was.  It’s easy for me to go back and chang warnings to errors, I’ll post on Slack now.

lrennels: @ckingdon95 for these I could add a helpful error message, but we probably still don't want to export these in the main API right? That said if I don't export them, users tried to use them without `Mimi.` prefix, then they're going to get an error anyways.

docs/src/tutorials/tutorial_5.md
```diff
@@ -351,7 +351,7 @@ my_scc_calculation (generic function with 1 method)
 
 Now that we have our post-trial function, we can proceed to obtain our two models and run the simulation. Note that we are using a Mimi MarginalModel `mm` from MimiDICE2010, which is a Mimi object that holds both the base model and the model with the additional pulse of emissions.
 
-```jldoctest tutorial4; output = false, filter = r".*"s
+```julia
```
lrennels, line 354: removed the dockets for now because MimiDICE2010 uses .marginal instead of .modified, can re-add it once the correct version of that model is used

lrennels:
NOTE to Lisa change the `replace!` function

davidanthoff:
Next step here is that @ckingdon95 approves and then we merge, right?

src/core/types/model.jl
```diff
@@ -45,11 +45,3 @@ end
...
     return (mm.modified[comp_name, name] .- mm.base[comp_name, name]) ./ mm.delta
 end
-
-function Base.getproperty(base::MarginalModel, s::Symbol)
-    if (s == :marginal)
-        @warn("Use of 'MarginalModel.marginal' will be deprecated, in favor of 'MarginalModel.modified'");
-        return getfield(base, :modified);
-    end
-    return getfield(base, s);
-end
```
ckingdon95, line 55: why did this get moved into the "model" file?

test/test_timesteparrays.jl
```diff
@@ -97,12 +97,6 @@ x[t3] = temp_dim_val[1]
 @test x[t3] == temp_dim_val[1]
 reset_time_val(x, time_dim_val)
 
-# Deprecated int indexing should still run
-@test x[3] == time_dim_val[3]
-x[TimestepIndex(3)] = temp_dim_val[3]
-@test x[TimestepIndex(3)] == temp_dim_val[3]
-reset_time_val(x, time_dim_val)
-
```
ckingdon95, line 105: should we leave part of this test to test that this now errors? (might help with our coverage of the functions we have deprecated, but the code is still there?)

ckingdon95: I guess this is a broader question about coverage of deprecated functions. @davidanthoff is there a way for us to list certain functions that the coverage tests don't care about? because this PR is definitely going to decrease our testing coverage unfortunately

lrennels: Yea, it is going to decrease coverage.  The other option I guess is to add tests for errors and then remove them when we remove the specific error messages in a later release I just wasn't sure if that was overkill and wanted to keep testing clean ... but coverage then goes down.

src/core/types/model.jl
lrennels: that must have been a mistake I'll move it back now!

ckingdon95:
@lrennels I noticed some tests that are failing, I think because of actual changes. I'm going to go through and address them now!

lrennels:
@ckingdon95 I think I just fixed one of them one second I'll push!

src/core/types/model.jl
```diff
@@ -46,9 +46,10 @@ function Base.getindex(mm::MarginalModel, comp_name::Symbol, name::Symbol)
     return (mm.modified[comp_name, name] .- mm.base[comp_name, name]) ./ mm.delta
 end
 
-# DEPRECATION - EVENTUALLY REMOVE
+# DEPRECATION - EVENTUALLY REMOVE (and go back to default getproperty behavior)
 function Base.getproperty(base::MarginalModel, s::Symbol)
     if (s == :marginal)
         error("Use of `MarginalModel.marginal` is deprecated in favor of `MarginalModel.modified`.")
     end
+    return getfield(base, s);
```
ckingdon95, line 54: ah thanks my bad for removing that!

lrennels: oh no worries I thought it was me :)

test/test_timesteparrays.jl
davidanthoff: `@test_throws` seems the right approach for this kind of situation?

ckingdon95: @davidanthoff I added back in the tests for old behavior that were already there, now with `@test_throws`. Now _overall_ test coverage for Mimi has increased, but the _diff_ coverage is not 100% because there were some deprecated Int-indexing functions that were not previously being tested, and that this PR changed from warnings to errors. Do you want me to add individual tests for each of these, or should we ignore it?

davidanthoff: Lets ignore that

davidanthoff:
This one is good now? @ckingdon95 can you approve?

PR #719, opened 2020-06-23
Move explorer to use React

davidanthoff:
@arnavgautam I think we're close here! I did two main things, 

(1) added a bunch of CSS to our work to mesh with what Material UI does (I put some links in the top of the CSS file for notes on where I found info)
(2) did a very preliminary Material UI theme adoption to get the font size down (https://codesandbox.io/s/l48vjmk3om?file=/index.js; https://codesandbox.io/s/l48vjmk3om?file=/index.js) but this could probably be a lot more elegant :)

Some potential TODOs:

- [x] make the data view list item go blue or orange when selected (we'd have to change them from list items to buttons or something (probably not worth too much effort on this first pass we can just make it a TODO for later especially since the graph title makes it clear what is selected)

- [ ] decide if we want the scroll bars to be always visible (that's the current state, see top of the CSS file)

- [ ] alter Material UI in a more elegant way ... although I don't really feel motivated to do this right now since it looks good as is ...

- [ ] any other changes you think would make it look better! We can show it at the meeting and see if people want something different!


src/core/defs.jl
```diff
@@ -99,6 +99,7 @@ istype(T::DataType) = (pair -> pair.second isa T)
 
 # Namespace filter functions return dicts of values for the given type.
 # N.B. only composites hold other comps in the namespace.
+components(model::Model, comp_name::Symbol) = components(compdef(model, comp_name))
 components(obj::AbstractCompositeComponentDef) = filter(istype(AbstractComponentDef), obj.namespace)
```
lrennels, line 103: I'm thinking that for this it might be good to open it up so that it acts on `obj::AbstractComponentDef` and then if it's called on a leaf component, which is a `ComponentDef`, it returns an empty array. In the future this seems like a cleaner way to handle recursive type calls ... @ckingdon95 thoughts on that?

src/core/model.jl
```diff
@@ -262,7 +262,17 @@ datumdef(m::Model, comp_name::Symbol, item::Symbol) = datumdef(compdef(m.md, com
 Return the dimension names for the variable or parameter `datum_name`
 in the given component `comp_name` in model `m`.
 """
-dim_names(m::Model, comp_name::Symbol, datum_name::Symbol) = dim_names(compdef(m, comp_name), datum_name)
+function dim_names(m::Model, comp_name::Symbol, datum_name::Symbol)
+    try
+        return dim_names(compdef(m, comp_name), datum_name)
+    catch e
```
lrennels, line 268: I'm guessing that the error here occurs if `comp_def` isn't in the top component?  I'm thinking it would be better to have an if-then rather than a try-catch just for clarity. In fact perhaps we just take out the shortcut for now, the second loop should work for any comp_name.

lrennels (commented):
@arnavgautam this is looking great!  I made a few comments and will now push a PR addressing them.  I may turn this into a non-draft PR now so that I can see the code coverage and add tests where necessary.

lrennels:
NOTE to self: as is expected at this point, there are tests in `test_explorer_model` and `test_explorer_composite_comp` that are now failing, probably because they are unit-testing functions that have now been changed.  A next step is to take a look at these two files and update them accordingly with the changes to the internal functions.

lrennels (approved):

Issue #717, opened 2020-06-18
finish v1.0.0 docs


lrennels:
Documentation required for release of v0.10.0 (and v1.0.0) marked with **TODO**
**1. Home** - a single file

**2. Tutorials** - header

-  Tutorials Intro 
-  Tutorial 1: Install Mimi
-  Tutorial 2: Run an Existing Model
-  Tutorial 3: Modify an Existing Model 
-  Tutorial 4: Create a Model 
-  Tutorial 5: Sensitivity Analysis

**3. How-to Guides** - header
- How-to Guides Intro
- How-to Guide 1: Construct and Run a Model 
- How-to Guide 2: View and Explore Model Results
- How-to Guide 3: Conduct Sensitivity Analysis 
- How-to Guide 4: Work with Timesteps, Parameters, and Variables
- How-to Guide 5: Port to Mimi v0.5.0
- How-to Guide 6: Port to Mimi v1.0.0 - **TODO** complete this

**4. Advanced How-to Guides** - header
- Advanced How-to Guides Intro
- Build and Init Functions
- Using Datum References

**5. Reference Guides** - header
- Reference Guides Intro
- Mimi API 
- Structures 1. Overview - **TODO** review and edit
- Structures 2. Definitions - **TODO** review and edit
- Structures 3. Instances - **TODO** review and edit

**6. FAQ**  - a single file

lrennels:
#724

PR #724, opened 2020-06-27
Documentation

lrennels:
This PR should address the minimal improvements to documentation for v0.10.0 and v1.0.0 to be released:

1. All published documentation should be correct
2. The integration guide for Mimi (>=) v0.5.0 to Mimi v1.0.0 should be complete

Handles #717


lrennels:
> I'm probably not the right person to review and approve, but what the heck :) I didn't check it carefully, but I think we should just merge it and move on!

Thanks! I think we might need you to override the merge PR, it's hanging on two tests for 1.2, x86 that aren't even going to run ... this shouldn't be an issue when I remove the 1.2 and 1.3 support for Mimi v1.0.0

PR #731, opened 2020-07-14
Change deprecation warning to error for sim instance access methods

lrennels:
This deprecation warning needs to be an error for v1.0.0, I noticed this when working on PAGE.

### DIFF
--- a/src/mcs/mcs_types.jl
+++ b/src/mcs/mcs_types.jl
@@ -230,6 +230,5 @@ function getdataframe(sim_inst::SimulationInstance, comp_name::Symbol, datum_nam
 end
 
 function Base.getindex(sim_inst::SimulationInstance, comp_name::Symbol, datum_name::Symbol; model::Int = 1)
-    msg = "getindex method for `SimulationInstance` has been replaced with getdataframe function for consistency of return type, please use getdataframe instead"
-    error("$msg")
+    error("getindex method for `SimulationInstance` has been replaced with getdataframe function for consistency of return type, please use getdataframe instead")
 end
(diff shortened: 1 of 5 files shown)


src/mcs/mcs_types.jl
```diff
@@ -231,6 +231,6 @@ end
 
 function Base.getindex(sim_inst::SimulationInstance, comp_name::Symbol, datum_name::Symbol; model::Int = 1)
     msg = "getindex method for `SimulationInstance` has been replaced with getdataframe function for consistency of return type, please use getdataframe instead"
-    Base.depwarn("$msg, $(stacktrace())", :getindex)
+    error("$msg, $(stacktrace())")
     return sim_inst.results[model][(comp_name, datum_name)]
```
davidanthoff, line 235: If this is an error, then we don't need the `stacktrace()` part (that will come automatically), and we also don't need the `return` statement, because it will never be reached.

lrennels: Good point, was clearly operating with 1/2 my brain there.

ckingdon95: hmm I think we are still using this stacktrace error message in the other functions we deprecated, should we remove it from all of them now too?

src/mcs/mcs_types.jl
```diff
@@ -231,6 +231,5 @@ end
 
 function Base.getindex(sim_inst::SimulationInstance, comp_name::Symbol, datum_name::Symbol; model::Int = 1)
     msg = "getindex method for `SimulationInstance` has been replaced with getdataframe function for consistency of return type, please use getdataframe instead"
-    Base.depwarn("$msg, $(stacktrace())", :getindex)
-    return sim_inst.results[model][(comp_name, datum_name)]
+    error("$msg")
```
davidanthoff, line 234:
```suggestion
    error(msg)
```
:) Or don't use a local var at all and just pass the string directly to `error`.

lrennels: All relics of older more complex messages from depreciations :) ... I’ll go through these today

PR #742, opened 2020-08-08
GlobalSA and Sobol Fixes

lrennels:
- introduces new `analyze` keyword arguments
- updates GlobalSensitivityAnalysis compat due to variance error in that package
- clarifies docs and adds tests

@ckingdon95 this is the first package that needs be updated with the following minimum version:
```
[compat]
GlobalSensitivityAnalysis = "1.0"
```
Once we merge this we should be able to update `MimiIWG` (I don't have permission so maybe you can do that, or I can do it if you give me permissions) and then finally we can update the repo for the paper. I think that's the required order due to dependencies.


test/mcs/test_defmcs_sobol.jl
```diff
@@ -68,6 +68,7 @@ results_disk[!,2] = Symbol.(results_disk[!,2])
 
 # do some analysis
 E = CSVFiles.load(joinpath(output_dir, "emissions_E.csv")) |> DataFrame
+results = analyze(si, E[1:60:end, 3]; progress_meter = false, N_override = 100)
 results = analyze(si, E[1:60:end, 3]; num_resamples = 10_000, conf_level = 0.95)
```
ckingdon95, line 72: Do we want to use `progress_meter=false` everywhere in the tests that use `analyze`? I think David said he prefers that

ckingdon95: also why are there two calls here? is it for test coverage? is it not possible to hit all the lines with one call because of the `N_override` if statements or something?

lrennels: Yes you're right we should make it false in all those cases.  I'm not sure off the top of my head maybe I use the `results` from line 72 in a later test?

ckingdon95:
> Once we merge this we should be able to update `MimiIWG` (I don't have permission so maybe you can do that, or I can do it if you give me permissions) and then finally we can update the repo for the paper. I think that's the required order due to dependencies.

Also, I think I don't need to explicitly update the compat in MimiIWG yet, I think just in the IWGReplicationPaper that specifically uses this... could be wrong though

test/mcs/test_defmcs_sobol.jl
lrennels: I guess I don't, so think I just wanted to check out both branches of setting N.

lrennels:
Oh I see, maybe the problem was the Mimi version of MimiIWG.  Does MimiIWG depend explicitly on the GSA package though?

test/mcs/test_defmcs_sobol.jl
ckingdon95: cool makes sense!

Issue #753, opened 2020-09-09
update_param! dirties the model definition


davidanthoff:
When I do this:
```julia
julia> m = MimiFUND.get_model()
Mimi.Model
  Module: Mimi
  Components: 
    ComponentId(MimiFUND.scenariouncertainty)
    ComponentId(MimiFUND.population)
    ComponentId(MimiFUND.geography)
    ComponentId(MimiFUND.socioeconomic)
    ComponentId(MimiFUND.emissions)
    ComponentId(MimiFUND.climateco2cycle)
    ComponentId(MimiFUND.climatech4cycle)
    ComponentId(MimiFUND.climaten2ocycle)
    ComponentId(MimiFUND.climatesf6cycle)
    ComponentId(MimiFUND.climateforcing)
    ComponentId(MimiFUND.climatedynamics)
    ComponentId(MimiFUND.biodiversity)
    ComponentId(MimiFUND.climateregional)
    ComponentId(MimiFUND.ocean)
    ComponentId(MimiFUND.impactagriculture)
    ComponentId(MimiFUND.impactbiodiversity)
    ComponentId(MimiFUND.impactcardiovascularrespiratory)
    ComponentId(MimiFUND.impactcooling)
    ComponentId(MimiFUND.impactdiarrhoea)
    ComponentId(MimiFUND.impactextratropicalstorms)
    ComponentId(MimiFUND.impactforests)
    ComponentId(MimiFUND.impactheating)
    ComponentId(MimiFUND.impactvectorbornediseases)
    ComponentId(MimiFUND.impacttropicalstorms)
    ComponentId(MimiFUND.vslvmorb)
    ComponentId(MimiFUND.impactdeathmorbidity)
    ComponentId(MimiFUND.impactwaterresources)
    ComponentId(MimiFUND.impactsealevelrise)
    ComponentId(MimiFUND.impactaggregation)
  Built: false


julia> m = MimiFUND.get_model()^C

julia> m.md.dirty
true

julia> run(m)

julia> m.md.dirty
false

julia> update_param!(m, :co2pre, 270)
true

julia> m.md.dirty
true
```
What this means is that code like this:
```julia
for i=1:10
  update_param!(m, :foo, rand())
  run(m)
end
```
will rebuild the model instance in every loop.

That is not how it should be, right? We want to be able to update external parameter values and rerun without rebuilding, it seems to me?

Or do we need to manually construct a model instance in that case and then update the parameter values only in the model instance?

[...]

PR #756, opened 2020-09-09
Cleanup build

davidanthoff:
The `build` function we previously had in the codebase doesn't return anything, but instead just modifies the model that is passed. So as a mutating function, it should really be named `build!`. First commit in this PR here makes that rename. I think that should be safe, given that this is an internal function.

The second commit then adds a `build` function that returns a model instance from a `Model`, but doesn't modify the `Model` itself. That is essentially what is documented in https://www.mimiframework.org/Mimi.jl/stable/howto_advanced/howto_adv_buildinit/#The-Internal-'build'-Function-and-Model-Instances-1, but not implemented currently.

This is a first step to address https://github.com/mimiframework/Mimi.jl/issues/753. What we then still need is an `update_param!` that works purely on a `ModelInstance`, and I think then we should be good?

### DIFF
--- a/src/core/build.jl
+++ b/src/core/build.jl
@@ -336,16 +336,21 @@ function _build(md::ModelDef)
     return mi
 end
 
-function build!(m::Model)
-
+function build(m::Model)    
     # Reference a copy in the ModelInstance to avoid changes underfoot
     md = deepcopy(m.md)
     _set_defaults!(md)  # apply defaults to unset parameters in the model instance's copy of the model definition
     
-    m.mi = _build(md)
+    mi = _build(md)
+    return mi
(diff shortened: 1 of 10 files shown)


lrennels:
@davidanthoff meaning that the `update_param!` is the "unsafe" version that @ckingdon95 mentions in #753 for users who know what they're doing and want to override?

lrennels (approved):

ckingdon95 (approved):

davidanthoff:
@lrennels I think eventually I'd like us to end up in a situation where one can update model parameters in a model instance, so that would solve the scenario but not be unsafe.

Issue #754, opened 2020-09-09
calling Packages within @defcomp


lrennels:
handle this: https://forum.mimiframework.org/t/set-a-parameter-as-a-distribution/106/15

PR #757, opened 2020-09-10
Parameterization with Package (ie. Distributions)

lrennels:
This handles #754, originally pointed out on the forum at https://forum.mimiframework.org/t/set-a-parameter-as-a-distribution/106/15.

We need to handle the issue of parameterizing a `Parameter` or `Distribution` with a type from another package like `Distributions`, specifically something like this:
```
lws_distribution = Parameter{Distributions.Normal{Float64}}()
```
The error originally stems from like 271 of `@defcomp`:
```
datum_type = (datum_type === nothing ? Number : Main.eval(datum_type))
```
@davidanthoff suggests the following : "We can discuss the problem with Distributions  on Fri. We might just have slightly rewrite the expression that gets passed to the Main.eval call, so that in the expression tree we don't use a symbol Distributions (which it will then try to resolve in the Main module, where Distributions isn't loaded), but instead just put a ref to the Distributions module itself into the AST that we pass to eval. I think it should then work without a name resolution."


lrennels:
This all makes sense to me and can take lead on this, but was wondering @ckingdon95 if you understand where in the `@defcomp` macro `datum_type` gets originally defined?  Is it implicitly created as a variable by one of the MacroTools macros? There's no explicit check for parameterization with another package so I think I'll need to add an `if-then-else` block to handle these cases or something.

lrennels:
@davidanthoff the calls look like this:

┌ Debug: before: datum type is Distributions.Normal{Float64}, which is of type Expr
└ @ Mimi ~/.julia/dev/Mimi/src/core/defcomp.jl:277
```
datum_type = (datum_type === nothing ? Number : Main.eval(datum_type))
```
┌ Debug: after: datum type is Normal{Float64}, which is of type DataType
└ @ Mimi ~/.julia/dev/Mimi/src/core/defcomp.jl:279

lrennels:
@davidanthoff ok I think this works ...

davidanthoff (approved):
Looks good to me.

Issue #748, opened 2020-08-22
Combined timestep and range indexing doesn't work


davidanthoff:
I tried this:
```julia
v.foo[TimestepIndex(1),:] .= 0
```
with this declaration:
```julia
foo= Variable(index=[time,4])
```
and that doesn't seem to work.

For now not a big problem because we can just do it manually, but it would be nice if this worked.

lrennels:
@davidanthoff I'll work on this!

lrennels:
Ah it looks like a `setindex!` problem, the `getindex` works ok

PR #758, opened 2020-09-10
Review Treatment of TimestepArrays

lrennels:
this PR is to work on #748 

It is fixed for the specific case @davidanthoff pointed out but really should be expanded to handle some other indexing cases ... I'm hoping there's a simpler way than writing all these cases into `dotview` but having trouble finding documentation on what `dotview` calls

**update** we should pause now and take a closer look at how we could possible remove any overrides of `dot view` in favor for a more conventional approach of adding `axes` in conjunction with `getindex` and `setindex!` methods ... and understand better the conventional way of doing this work


lrennels:
This will fix the specific issue cited ... I'll add an issue to make this more robust in the future.

davidanthoff (approved):
Ok, lets go with this for now and then review our general approach later.

Issue #760, opened 2020-09-11
Remove remaining eval call in the @defcomp


davidanthoff:
I think we should be able to get rid of the remaining `eval` call in `defcomp` as well, [here](https://github.com/mimiframework/Mimi.jl/blob/2538cd288c48263f461dfaae8c62aef13740e29e/src/core/defcomp.jl#L265).

I think all we would have to do is move the check that is there into `add_parameter` [here](https://github.com/mimiframework/Mimi.jl/blob/2538cd288c48263f461dfaae8c62aef13740e29e/src/core/defcomp.jl#L85) so that it doesn't occur during macro expand time but rather when the code that is generated by the macro actually runs. And then in the AST we just use the same strategy that we now used for the types, i.e. we just put the expression into the AST instead of the evaled expression.

That should in general all be way more stable that way, because a) the expression for the default value will then be evaluated in the module where it is written down and b) in general it shouldn't be necessary (or advised) to call `eval` inside a macro.

@lrennels ok if I assign this to you as a follow up of the previous one?

PR #761, opened 2020-09-12
Remove eval from defcomp macro

lrennels:
#760

### DIFF
--- a/src/core/defcomp.jl
+++ b/src/core/defcomp.jl
@@ -82,6 +82,11 @@ function add_variable(comp_id::ComponentId, name, datatype, dimensions, descript
 end
 
 function add_parameter(comp_def::ComponentDef, name, datatype, dimensions, description, unit, default)
+    if default !== nothing
+        if length(dimensions) != ndims(default)
+            error("Default value has different number of dimensions ($(ndims(dflt))) than parameter '$name' ($(length(dimensions)))")
+        end
+    end
     p = ParameterDef(name, comp_def.comp_path, datatype, dimensions, description, unit, default)
     comp_def[name] = p            # adds to namespace and checks for duplicate
     dirty!(comp_def)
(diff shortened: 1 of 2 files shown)


davidanthoff (approved):
Excellent!

ckingdon95:
sorry I'm late to seeing this after you merged it-- but I see that you had to change some of the tests for defaults. Did you make sure that the changed behavior makes sense? Cause it looks like you moved the code for checking the number of dimensions for a default value, and codecov says that those modified lines aren't being hit by the tests, which is supposed to be the test that you modified. does that make sense?

src/core/defcomp.jl
```diff
@@ -82,6 +82,11 @@ function add_variable(comp_id::ComponentId, name, datatype, dimensions, descript
 end
 
 function add_parameter(comp_def::ComponentDef, name, datatype, dimensions, description, unit, default)
+    if default !== nothing
+        if length(dimensions) != ndims(default)
+            error("Default value has different number of dimensions ($(ndims(dflt))) than parameter '$name' ($(length(dimensions)))")
+        end
```
ckingdon95, line 88: ah actually it is just the "end" line of the if statement that codecov is complaining about. But I was surprised that the error type for that error would be `UndefVarError` and not still a `LoadError`, so I think there might be a problem here that `dflt` isn't defined here and that it should be `default`?

davidanthoff: Ah, yes, good catch! The test is now reporting a bug in the line that throws the error, i.e. in line 87 :) The test should actually test for an `ErrorException`, right? That is what is thrown, so any other type of error would mean something went wrong.

lrennels: Good point, I was lazy when I changed that, I’ll change the error type back and correct dflt to default, I think that should solve it.

lrennels: Sorry, but did we conclude that it should be a `LoadError` or `ErrorException`?  It's still an `ErrorException` (see ), but even with `default` instead of `dflt`. Is there a problem here because we're calling `ndims` on `default` before it is evaluated ... or has it been evaluated at this point?

ckingdon95: I think David is right, it should now be `ErrorException` (it was previously `LoadException` when it was being thrown from within eval I think)

ckingdon95: Also if you just run the part of the test file that's supposed to test this error, but run it outside of the `test_throws` then you can see if the error message is our constructed error message or if something else is wrong, just to be safe

lrennels: @ckingdon95 smart yes I just did that and the constructed error message looks correct, so I think we can merge the PR now if someone can approve it

PR #772, opened 2020-10-21
Call set_param! or update_param! for set_index

jrising:
Since the reference system abstracts away from `set_param!` and `update_param!`, we need to call whichever is appropriate. This means `set_param!` the first time a parameter is set and `update_param!` on subsequent times.

### DIFF
--- a/src/core/references.jl
+++ b/src/core/references.jl
@@ -2,20 +2,24 @@
     set_param!(ref::ComponentReference, name::Symbol, value)
 
 Set a component parameter as `set_param!(reference, name, value)`.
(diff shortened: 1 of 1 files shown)


jrising:
@lrennels This is a minor fix, so that I can do:
```
compref[:param] = value1
...
compref[:param] = value2
```
without provoking an error.

lrennels:
Hi @jrising just so I understand better, why isn't it possible in this case to use `set_param!` followed by `update_param!` later?

lrennels:
@jrising would #773 do what you want as well?  It might be preferable in terms of code cleanliness even though it's more verbose.  I'll add a test too.

PR #773, opened 2020-10-23
Fix set_param for references

lrennels:
@jrising would this work to fix your bug at #772  too?  I know it looks more verbose but it might be clearer in terms of code interpretability ...


ckingdon95:
If this works for James, I would prefer this way!

jrising:
@lrennels Yes, I definitely see the benefit of doing it this way. I'm happy with this change.

Thank you!

ckingdon95 (approved):

davidanthoff (approved):
Yes, that is the way to do it!

Issue #752, opened 2020-09-01
Should delete! on a component delete stray parameters?


davidanthoff:
I've run into the following scenario: someone gets a model via the `m = ModelName.get_model()` function. I then remove a component from `m` via the `delete!` function. At this point `m` now can have a number of external parameters that are not used anywhere. Say one of the left-over parameters if called `foo`. `foo` was used by the component that I just deleted.

Now I add a new component that also has a parameter `foo`. Of course when I call `set_param!(m, :compName, :foo, 3)` it now errors because `m` already has a parameter called `foo`.

I'm not sure what the right behavior here would be, and maybe the current one is actually correct....

But here are some alternative options:
1) `delete!` on a component could delete stray parameters like `foo`. That seems a bit drastic, though... Maybe we need a keyword argument `also_delete_stray_params` that one can use if one wants to get rid of stray parameters?
2) or is this really a scenario where the replace component functionality should be used?
3) do we have a function that deletes an external parameter from a model?

Also CCing @frankerrickson, what he thinks might be useful here.

davidanthoff:
@ckingdon95  and I decided to go with option 1. Just need a good name for the keyword argument. `delete_unused_model_params` was one idea.

PR #774, opened 2020-10-23
add `delete_unbound_comp_params` kwarg to `delete!`

ckingdon95:
#752 

Two additions in this PR:
1. This adds `delete_unbound_comp_params::Bool=false` keyword argument to the `delete!` function, which when set to true then deletes all external parameters that were only connected to the deleted component. 
2. I also created a new function `delete_param!(m, external_param_name, delete_connections=true)` for removing parameters from a model's list of external parameters.

To discuss:

- kwarg name: I went with the verbose `delete_unbound_comp_params` for now, which is definitely not perfect. The reason I'm feeling averse to `deep` or `recursive` is that it doesn't delete _all_ parameters associated with this component, only the ones not connected to any other components.
- in order to add this functionality, I had to change the function signature to only operate on ModelDefs, and not on all AbstractCompositeComponentDefs because the latter do not have external parameter lists.
- for the new `delete_param!` function, is this okay to make the default `delete_connections=true`, and do we want a different name for that keyword arg?

### DIFF
--- a/docs/src/ref/ref_API.md
+++ b/docs/src/ref/ref_API.md
@@ -7,6 +7,7 @@ Model
 add_comp!  
 connect_param!
 create_marginal_model
+delete_param!
 dim_count
 dim_keys
 dim_key_dict
--- a/src/Mimi.jl
+++ b/src/Mimi.jl
@@ -19,6 +19,7 @@ export
     # components,
     connect_param!,
     create_marginal_model,
+    delete_param!,
     dim_count,
     dim_keys,
     dim_key_dict,
--- a/src/core/defs.jl
+++ b/src/core/defs.jl
@@ -96,6 +96,7 @@ function Base.delete!(md::ModelDef, comp_name::Symbol; delete_unbound_comp_param
         epc_filter = x -> x.comp_path != comp_path
         filter!(epc_filter, md.external_param_conns)
     end
+    dirty!(md)
 end
 
 """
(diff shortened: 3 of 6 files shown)


davidanthoff:
For the new `delete_param!` function, I don't fully understand what `delete_connections=true` does. Don't we always have to delete connections that included the deleted parameter, because otherwise we would have "stray" connections hanging around?

ckingdon95:
@davidanthoff Yes, it would be a weird behavior to leave the connections hanging. Shall I remove the option all together for now? (Since this is new, we could always add in options later if we realize they might be needed)

ckingdon95:
@davidanthoff ok I got impatient and was assuming you were going to answer yes so I removed that option for now

ckingdon95:
One design question is whether it would be better to not have a separate `delete_param!` function and if this functionality should be folded into the `delete!` function. basically it would be:
```
function delete!(md::ModelDef, item_name::Symbol)
    item = md.namespace[item_name]
    if item isa component
        _delete_component . . .
    elseif item isa param
        _delete_param . . .
    end
end
```
The only thing is that we now have this `delete_unbound_comp_params` option that only makes sense in the component case

ckingdon95:
decided in group meeting: stick with `delete_param!` because we don't have a shared namespace for components and parameters

davidanthoff (approved):

PR #775, opened 2020-10-23
Add temporary Model constructor

lrennels:
@davidanthoff  here's a temporary constructor for @FrankErrickson ... should work for now can you approve?


davidanthoff (dismissed):
Should we also add the `explore(mi::ModelInstance)` method right away? Then users like @frankerickson can use that, which we do want to support long-term, and don't need to take a dep on this constructor?

lrennels:
@davidanthoff sure, you mean add `explore(mi::ModelInstance)` and then have that call `explore(Model(mi))`?  Yea that's smart then the API won't change when we fix it.

davidanthoff (approved):
Perfect!

Issue #779, opened 2020-11-02
Monte Carlo Broadcasting Bug


lrennels:
Per Dale's comment here:

https://forum.mimiframework.org/t/mcs-assigning-random-variables-to-indexed-variable/129

due to changes in broadcasting etc. in underlying Julia, instructions here: https://forum.mimiframework.org/t/mcs-assigning-random-variables-to-indexed-variable/129 for assigning to slices no longer work when you are doing a many to one assignment and thus need a `.=` operator.  We have added the `.=` operator into the code in a PR, but need to add underlying array functionality as pointed out in #765 for this to work ...

we may be able to hack it within `_perturn_param!`'s  `.=` clause but not sure how just yet

PR #780, opened 2020-11-04
Broadcast

lrennels:
This PR handles the issue shown by #779 in the underlying function `_perturb_param!` called within `@defsim`.  Broadly the issue arises when there is a need to use broadcasting.  This is problematic because (1) broadcast assignment of `TimestepIndex` to `TimestepArray` does not have the proper underlying methods and (2) it is not always clear whether there is a need for `.=` or `=` even when there is no time index.

The current solution control-flow can be described as follows.  It is more complex/deep than I'd like, but it's the simplest I have for now.  **NOTE FLIP Y AND N ON HAS TIME INDEX FORK**

image


lrennels:
@ckingdon95 I worked on this today and removed a lot of the control flow and the try/catch.  It should work properly now, could you take a look at the changes and see if you think there is a better way to simplify or move common cases out of loops etc.?  It would be good to get some fresh eyes on it :)

lrennels:
@ckingdon95 the doctests are being dumb right now so I removed them, but I added an Issue to add them back as soon as I can! If this looks good we can merge it and tag a new version soon. I'll keep a close eye on performance going forward as well.

docs/src/tutorials/tutorial_3.md
```diff
@@ -54,13 +54,10 @@ You have now successfully downloaded MimiDICE2010 to your local machine.
 
 The next step is to run DICE using the provided API for the package:
 
-```jldoctest tutorial2; output = false, filter = r".*"s
+```julai
```
arnavgautam, line 57: "julia"?

arnavgautam (dismissed):

docs/src/tutorials/tutorial_3.md
lrennels: oh shoot yes thanks, let me make sure the tests all pass then I'll ping you again

arnavgautam (approved):

Issue #573, opened 2019-10-11
Reenable first and last functionality


davidanthoff:
Maybe also revisit the design.

lrennels:
#793

PR #793, opened 2021-01-28
Time dims

lrennels:
This PR should handle issues #573 and #422 with the main goal of reenabling the `first` and `last` keyword arguments on `add_comp!` so that we can use components of different time lengths.


lrennels:
From conversation 1/28 with David:
- For now, we will enforce that when you `set_param!` or `update_param!` the dimensions of the data must exactly match the dimensions of the model at the time.  Note that if a dimension has been updated with `set_dimension!`, this may mean that the new parameter values have a different size than the old ones.  Also note that this means that parameters for components with shorter time dims than the model will needed to be padded with default values or `missing`s.  We will try to relax this later.

lrennels:
The general story right now is:
All provided data (with set_param! or update_param! must match the full dimensions of the model.  You can pad with missings if you want, but the dimensions have to match even for :time.  This is a change, but it really just clarifies previously buggy behavior.  We will try to loosen this later!
Each component def, noting that a ModelDef is a type of component def, holds both a :time dimension and a first and last.  The :time dimension of all components at the time of a model running are the same, specific operations like set_dimension! and build! propagate the top-level time dimension keys through to ensure that fact. If first and last are not set explicitly, they will be at the bounds of the model :time dimension and change with the model if the dimension is reset with set_dimension!.  If they are set explicitly, they are fixed and immutable, and apply to all subcomponents of the component on which they were fixed during add_comp!. 
The model can start before any component, and end after, you’ll just get missings in your output arrays.  A note here is that if you used something like is_first(t) in your run_timestep to set an initial value… you might run into problems because that refers to the model time, not the compoent time.  That can be fixed by using something like if t == TimestepValue(2020) which I tried out.

lrennels:
Update: going to alter so that `is_first` refers to the first tilmestep a component instance is run, not always the model tilmestep

lrennels:
From experimenting with coupling FAIR and FUND, I've realized an inconsistency with `TimestepIndex`.  While comparison operators like `if t == TimestepIndex(5)` works correctly because it compares `ts.t` with `5` ... indexing into a `TimstepArray` with `TimestepIndex(5)` isn't going to work because there simply isn't the information we need to offset the `5` by the component since our storage story has the data the length of `time`.  @davidanthoff @ckingdon95 we can discuss

arnavgautam (approved):

Issue #795, opened 2021-02-17
MCS modification functions


lrennels:
- write some documentation for the MCS modification functions
- make sure all MCS modification functions have proper errors if a RV is not found ex. add an error if `:replace_RV` is called for a parameter that does not exist, right now it adds one which is confusing and unexpected behavior
- write a test file for MCS modification functions

- explicitly note that if you don't create names for your random variables automatic ones are created and maybe nondeterministic ... good idea to at some point add a function to do something like ... 
```
mcs = getsim()
replace_RV!(mcs, :name, Normal(1,2)
```
when :name was defined with the shorthand
```
name = Normal(0,1)
```
and thus has a automatically generated key in `mcs.rvdict` in the format `:name!number`

PR #796, opened 2021-02-17
RV modification functions

lrennels:
#795


lrennels:
@grantmcdermott this should help with some of the issues you ran into here: https://github.com/anthofflab/MimiPAGE2009.jl/issues/222

grantmcdermott:
Super, thanks for the head up @lrennels!

Two minor aside, which I hope you don't mind me mentioning here although they're only tangentially related to this PR.
- Assuming this is still the case, could you mention somewhere in the documentation that `EmpiricalDistribution` isn't export by default when loading Mimi?
- Talking documentation, is there a reason why "Documentation" isn't a main sub-menu link on the left-hand side of the Mimi [website](https://www.mimiframework.org/)?

lrennels:
@grantmcdermott I don't mind at all, feedback is really helpful. Per your first comment yes that's a good idea I'll add some documentation on the status of `EmpiricalDistribution`, need to go look at if there's a reason we haven't documented it (i.e. it's in Beta format and we don't want to throw users in the deep end unless they're comfortable with it), or we just hadn't done it yet.  I think I didn't put it in the sub-menu link since those are mostly links to pages within the forum as opposed to external links, but if you think it would be helpful I could put it there!  I'm not sure the left hand side label can be a link, but it could link to a page that provides the link.

lrennels:
@grantmcdermott just added a documentation link to the left-side bar, good idea!

wip/temp.jl
```diff
@@ -0,0 +1,34 @@
+using Mimi
```
arnavgautam, line 1: Was this file intended to be checked in? I can't see how it connects to the rest of the files in this PR

arnavgautam (dismissed):
Always exciting to see us surfacing more details on errors

wip/temp.jl
lrennels: Ah good point thank you, deleted it!

arnavgautam (approved):

PR #800, opened 2021-03-09
fancy functions for set_param! (fix #524)

### LINKED ISSUE #524: Changes to set_param!

wingfactor:


wingfactor:
#524

lrennels:
Hi @wingfactor thank you for pushing forward on these issues!  On Monday we're going to have a broader team meeting on some issues with `set` and `update` param calls that have come up with users, so that might influence some of the codebase you are working on.  Regardless it's a good exercise in getting pretty deep into the code for you to work on these, so by all means continue, just a warning that you and I may want to work together in the coming weeks if I start messing with the same code :)

wingfactor:
Hi @lrennels Thank you for letting me know about that. I am definitely on board for collaborating on the v1.2.0 milestone issues!

What do you think about #524 ? It seems to me that once the code from David's initial comments runs properly, it might be ready to be closed.

lrennels:
You’re probably right, I haven’t looked at that Issue in a long time so I’m not sure how it folds in to current behavior. 

I think our first discussion is mostly going to be high level, and addressing a problem where setting default values means you have to use `uodate_param!` even though the user hasn’t seen it explicitly set yet ... that’s confusing a lot of users. 

I can be in touch after our meeting, but feel free to request my review on this sooner or ask me questions, we can fold it in before I make changes if we think that makes sense.

wingfactor:
I'll reopen this once it's closer to being finished.

Issue #651, opened 2020-02-05
improve explorer for non-time variables


ckingdon95:
```julia
using Mimi

@defcomp example begin
    p1 = Parameter(index = [foo])
    p2 = Parameter(index = [foo, bar])
    p3 = Parameter(index = [foo, baz])
    p4 = Parameter(index = [time, bar])
    p5 = Parameter(index = [time, baz])
end

m = Model()
set_dimension!(m, :time, 10)
add_comp!(m, example)
set_dimension!(m, :foo, 1:5)
set_dimension!(m, :bar, 1:3)
set_dimension!(m, :baz, [:A, :B, :C])
set_param!(m, :example, :p1, 1:5)
set_param!(m, :example, :p2, reshape(1:15, 5, 3))
set_param!(m, :example, :p3, reshape(1:15, 5, 3))
set_param!(m, :example, :p4, reshape(1:30, 10, 3))
set_param!(m, :example, :p5, reshape(1:30, 10, 3))
run(m)
explore(m)
```
I think the plot for p1 should be with points instead of bars. (If the dimension values are numerical, I think they should get plotted as a scatter plot.)

Then the plots for p2 and p3 are currently meaningless, but what's interesting is that they are rendered differently from each other I guess because one of them has symbols for the second dimension and one of them has numbers for the second dimension. I think we should stick with putting the first dimension, foo, as the x axis, and then still produce a scatter plot where the second dimension is represented by different colors, like we do for p4 and p5.

There's probably more corner cases we need to define, but this is a start!

rjplevin:
```
set_param!(m, :example, :p1, 1:5)
set_param!(m, :example, :p2, reshape(1:15, 5, 3))
```
In the current code under development, the first statement won't work because the import will not have happened yet. We might import on demand if the requested param doesn't exist in `m.md` and it is unbound in one subcomponent (but not in multiple). 

Alternatively, you could call `import_params!(m)` any time after the `add_comp!` and before the `set_param!` calls. Alternatively alternatively, we could add a kwarg, e.g., `add_comp!(m, import_all=True)`.

ckingdon95:
I'm pretty sure we want those calls to `set_param!` to still work. But I think any discussion about that belongs on this issue: https://github.com/mimiframework/Mimi.jl/issues/524

PR #803, opened 2021-03-16
Improve Explorer for Non-Time Data

lrennels:
#651


arnavgautam (approved):
My only comment is that getting the JSONText for the different plot types seems somewhat repetitively, but I also can't see a better way of doing it, so this looks good!

lrennels:
@arnavgautam I agree that really should be revisited!

Issue #805, opened 2021-04-01
Change Component first/last after it is added


lrennels:
Right now when we add a component to a model, we can optionally set its `first` and `last`.  If they are set there are set the component's `first` and `last` **will not change** if the model time dimension is changed (they have a `first_free` and `last_free` flag that is set to `false`.  If they are not set, they will take the first and last of the model itself, and **vary freely** with the model, so if someone calls `set_dimension(:time, values` on the model, that time will propagate through the components and change the `first` and `last` of the components.

This creates a problem in the FUND_FAIR case:

We start with 
```
m = MimiFUND.get_model(;nsteps = 551 )
```
Currently, `m` runs from 1950 to 2500, and the components and model all have first and last of 1950 and 2500 respectively.  Now we want to widen the time dimension so that we can couple FAIR, so we call:
```
set_dimension!(m, :time, collect(1765:2500))
add_comp!(m, MimiFAIR_composite, after = MimiFUND_emissionscomposite)
```
The problem now is that this time dimension automatically flows through all the FUND components, so all FUND components have a `first` and `last` of 1765 and 2500 respectively ... which is wrong! 

Thus we need a user-facing function to retroactively set `first` for the FUND components.  This should call my existing functions `Mimi.propagate_time!` (or `Mimi._propagate_firstlast!`) under the hood and will look something like this ...

`set_firstlast!(m::ModelDef, comp_name::Symbol; first = NothingInt, last = NothingInt)`

so that we can call
```
m = MimiFUND.get_model(;nsteps = 551 )
set_dimension!(m, :time, collect(1765:2500))
add_comp!(m, MimiFAIR_composite, after = MimiFUND_emissionscomposite)

set_firstlast!(m, MimiFUND_emissionscomposite, first = 1950);
set_firstlast!(m, MimiFUND_climatecomposite, first = 1950);
set_firstlast!(m, MimiFUND_impactscomposite, first = 1950);
```

Ideas for the name?  I guess it could be `update_firstlast!`? Also can we put this in the exported API or should I keep it out?

lrennels:
#811

PR #807, opened 2021-04-05
Firstlast

lrennels:
issue summarized here: #805 

* per the next few PRs coming through this might not really be necessary, but keep around the Draft until we decide!


lrennels:
TODO add check that first and last appear in the model dimension

Issue #560, opened 2019-10-01
Add other versions of `adder` component


ckingdon95:
We may want to add versions of `adder` that allow for `missing` values, or that allow `first` and `last` bounds.

examples (need better names):

```julia
@defcomp adder2 begin
    add    = Parameter(index=[time])
    input  = Parameter(index=[time])
    output = Variable(index=[time])

    function run_timestep(p, v, d, t)
        v.output[t] = Mimi.@allow_missing(p.input[t]) + p.add[t]
    end
end
```
or 
```julia
@defcomp adder3 begin
    add    = Parameter(index=[time])
    input  = Parameter(index=[time])
    output = Variable(index=[time])

    first::Int = Parameter(default=0)
    last::Int  = Parameter(default=typemax(Int))

    function run_timestep(p, v, d, t)
        if first < getindex(t) < last
            v.output[t] = p.input[t] + p.add[t]
        end
    end
end
```
Note that in this version of `adder3`, we would need to decide if `first` and `last` are index values, in which case we need to add a `getindex(t)` function, or if they refer to actual year values, in which case we would need to use `gettime(t)` and this would only work for time indexes that are Ints.

PR #816, opened 2021-04-23
Add multiplier and allow_missing macro use

lrennels:
This PR (1) adds a `multiplier` build in component to help with FUND-FAIR unit conversion and (2) per suggestion from #560 adds an `@allow_missing` wrapper to the `adder` and `multiplier` components, which also simplifies MimiFUND-MimiFAIR-Flat.

This means that if the multiplier component:

```
@defcomp multiplier begin

    multiply  = Parameter(index=[time])
    input       = Parameter(index=[time])
    output      = Variable(index=[time])

        v.output[t] = @allow_missing(p.input[t]) * p.multiply[t]
    end
end
```
accesses a `missing` in `p.input[t]` it will not throw an error but just pass the value through.  Sometimes needing this can be avoid with setting a `first` and `last` for the `multiplier` component, but sometimes there's still an irritating off-by-one issue. For example in MimiFUND-MimiFAIR-Fla we have
```
set_param!(m, :multiplier, :multiplier, 1/1000) # convert Mtons CO₂ coming out of FUND to Gtons CO₂ going into FAIR
connect_param!(m, :multiplier, :input, :emissions, :mco2)
connect_param!(m, :co2_cycle, :E_CO₂, :multiplier, :output, FAIR_CO₂_backup, backup_offset = 1)
```
where the `:mco2` from FUND's `:emission` is `missing` for the first year FUND runs (1950), and thus we would need `multiplier` to run from `FUND_first + 1` to `FUND_last` ... or do something fancy with backups which also works.  In my opinion we want these to be sort of no-brainer pass-through components that don't do anything fancy so I'd rather just do `@allow_missing` but we can discuss if you'd like.

### DIFF
--- a/docs/src/internals/structure.md
+++ b/docs/src/internals/structure.md
@@ -80,4 +80,4 @@ In the new version, all component definitions are represented by one type, `Comp
 
 ## 3. Pre-compilation and built-in components
 
-To get `__precompile__()` to work required moving the creation of "helper" components to an `__init__()` method in Mimi.jl, which is run automatically after Mimi loads. It defines the two "built-in" components, from `adder.jl`, `multiplier.jl`, and `connector.jl` in the `components` subdirectory.
+To get `__precompile__()` to work required moving the creation of "helper" components to an `__init__()` method in Mimi.jl, which is run automatically after Mimi loads. It defines the three "built-in" components, from `adder.jl`, `multiplier.jl`, and `connector.jl` in the `components` subdirectory.
--- a/docs/src/internals/structure.md
+++ b/docs/src/internals/structure.md
@@ -80,4 +80,4 @@ In the new version, all component definitions are represented by one type, `Comp
 
 ## 3. Pre-compilation and built-in components
 
-To get `__precompile__()` to work required moving the creation of "helper" components to an `__init__()` method in Mimi.jl, which is run automatically after Mimi loads. It defines the two "built-in" components, from `adder.jl` and `connector.jl` in the `components` subdirectory.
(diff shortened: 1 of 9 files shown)


src/components/README.md
```diff
@@ -13,5 +13,7 @@ module. So to place components defined here in `Mimi`, prefix the component name
 
 * `adder.jl` -- Defines `Mimi.adder`, which simply adds two parameters, `input` and `add` and stores the result in `output`.
 
+* `multiplier.jl` -- Defines `Mimi.multiplier`, which simply multiplies two parameters, `input` and `add` and stores the result in `output`.
```
arnavgautam, line 16: The second parameter here is 'multiply' right?

docs/src/internals/structure.md
```diff
@@ -80,4 +80,4 @@ In the new version, all component definitions are represented by one type, `Comp
 
 ## 3. Pre-compilation and built-in components
 
-To get `__precompile__()` to work required moving the creation of "helper" components to an `__init__()` method in Mimi.jl, which is run automatically after Mimi loads. It defines the two "built-in" components, from `adder.jl` and `connector.jl` in the `components` subdirectory.
+To get `__precompile__()` to work required moving the creation of "helper" components to an `__init__()` method in Mimi.jl, which is run automatically after Mimi loads. It defines the two "built-in" components, from `adder.jl`, `multiplier.jl`, and `connector.jl` in the `components` subdirectory.
```
arnavgautam, line 83: Are there three "built-in" components now?

arnavgautam (dismissed):
Just minor comments on the tutorial language

arnavgautam (approved):

Issue #812, opened 2021-04-20
Updates to set_param! and update_param!


lrennels:
Below I summarize three issues that this PR should encompass, all of which involve `update_param!` and `set_param!`.  

I do think at this point we should be careful about breaking things or adding too much API ... keeping things as fluid and simple for users as possible is a priority!

**1. `set_param!` vs `update_param!`**

There has been a little bit of dissatisfaction with having to decide between these two functions.  I can understand that, but don't see a very clean way to put them together given the current structure.  I guess we could add a function under the hood that does something like "if the parameter name is in the external parameter list, update it and connect the component, otherwise create a new one" ... but that first clause is dangerous because we normally error when you try to `set_param!` with an existing parameter.  See next issue.

**2. Dissaggregating `create_external_param` from `connect_param!` #750**

I would read [this discussion](https://github.com/anthofflab/MimiFAIR.jl/pull/19#pullrequestreview-472965223) to understand the issue, it's fairly concise.

The idea here would be that we expose the two steps contained within `set_param!` and use `create_param!` and `connect_param!` or something, instead of a using a `set_param!` call without specifying a component as a way to set the parameter for all of the components with that parameter name.  Or we could make it explicit with a call like 
```
set_param!(m, [:compA, :compC, :compD], :p1, 23)
``` 
instead of 
```
set_param!(m, :p1, 23)
```

**3.  Handling Defaults (#788)**

Current behavior:
```
using Mimi
using MimiFUND
m = MimiFUND.get_model()

# cannot update_param! for one with a default
update_param!(m, :climatesensitivity, 2.0) 
ERROR: Cannot update parameter; climatesensitivity not found in composite's external parameters.

# can set_param!
set_param!(m, :climatesensitivity, 2.0) # works! 
```
while this may make sense, because the model has not been built yet and we want users to be able to call `set_param!` just like everything else, **the same behavior remains even after we have called run(m)**, ie.
```
run(m)
# STILL cannot update_param! for one with a default
update_param!(m, :climate sensitivity, 2.0) 
ERROR: Cannot update parameter; climatesensitivity not found in composite's external parameters.
```
This strikes me as unexpected. Up until the model is run, it seems fair that the model acts like the default does not exist and the parameter needs to be `set`.  But once it's run and the defaults were applied once, users expect that a model that has run, and thus needed values for all parameters, they should be to call `update_param!` on any parameter.

[...]

PR #829, opened 2021-05-06
Simplify Parameter Handling

lrennels:
Handles a bunch of issues summarized in #812 

Corresponding necessary model updates:

- PAGE2009: anthofflab/MimiPAGE2009.jl#231
- PAGE2020: openmodels/MimiPAGE2020.jl#74
- MimiIWG:rffscghg/MimiIWG.jl#42
- MimiFUND: fund-model/MimiFUND.jl#99
- MimiFAIR-MimiFUND-Flat: TODO


lrennels:
Per discussion 5/12/2021:

1. `replace!` behavior 

Question: How do we want `replace!` to handle new defaults and new parameters?
Answer: When we replace a component with a new component, but choose to maintain all old connections, we will maintain the old default settings, and not replace them with the new component's defaults.  That said, we might need to add in some new `nothing` unshared external parameters to our external parameter list, so upon the call of `add_comp!` we should add those but not add any for parameters that are already connected.

2. composite defaults behavior

Question: How do we want to handle defaults from composites?
Answer:  We want to make sure that every single parameter gets an unshared external parameter, including all of the leaf component ones bubbled up and any new ones in the composites including ones that are simply renamed for collision fixes.  For those, we need to get the default value bubbled up as well, and then if people want to turn that off they can always use `default = nothing`.  That should be enough to get everything into the external parameter list but we really need to add tests for this behavior!

3. Monte Carlo

Question: How do we (1) not break old behavior where we gave a parameter distribution to a default value (and thus shared named external parameter) in the simulation definition and (2) handle unshared parameters and defaults goign forward.
Answer: We will add a special check for the `p = Normal(0,1)` syntax such that if we can't find `p` we'll go look to see if there is a unique component parameter with a default that we can use that for ... if not we can error and suggest the new syntax.  Our new syntax will be `comp.parameter = Normal(0.,1)` for non shared parameters.

@davidanthoff this is a rough summary of what we just discussed.  After I work through this and shore up some testing, I will go back to these next steps:

1. external API: add `update_param(m, comp, param, value` (for unshared parameters) and `create_model_parameter` (for shared parameters)
2. make sure our internals start making use of `update_param!` instead of `set_param!` so most things are unshared and we can phase out `set_param!`
3. add a bunch of tests, and figure out how to adapt old tests to make sure `set_param!` still works properly but `update_param!` is the more common use case.  One way to do this is mostly use existing models to test for `set_param!` and move testing towards `update_param!`.
4. documentation

lrennels:
Next Steps ...

@davidanthoff this is a rough summary of what we just discussed.  After I work through this and shore up some testing, I will go back to these next steps:

- [x] external API additions
- `update_param!(m, comp, param, value` and it's @delegate methods
- `add_shared_parameter` and it's @delegate methods
- [x] do some work getting rid of "external parameter" replaced with "shared model parameter" or "unshared model parameter"
- [ ] make sure our internals start making use of `update_param!` instead of `set_param!` so most things are unshared and we can phase out `set_param!`
- [ ] add a bunch of tests, and figure out how to adapt old tests to make sure `set_param!` still works properly but `update_param!` is the more common use case.  One way to do this is mostly use existing models to test for `set_param!` and move testing towards `update_param!`.
- [ ] documentation 
- [ ] `update_leftover_params!` - maybe a new function where you can batch update unshared parameters?

davidanthoff (approved):

PR #836, opened 2021-07-08
Empirical Distribution `probs` Vector Bug

lrennels:
This PR fixes bugs in the DataFrames syntax that prevented an EmpiricalDistribution type in a `@defsim` from working with explicitly provided probabilities. We also add a test.


lrennels:
we should also add a test!

davidanthoff (approved):

PR #839, opened 2021-07-14
Add warning for comp dot parameter defsim

lrennels:
This PR adds a warning if a user uses the `compname.paramname = Distribution` syntax for a component parameter linked to a shared parameter. This should be a warning, if not an error, because it may look as if the user is varying just that component's parameter, but this will vary the shared parameter and thus all parameters connected to that shared parameter. The TODOs here are

- [x] add the warning
- [x] test the warning


lrennels:
#838

davidanthoff (approved):

PR #842, opened 2021-09-08
Add set param methods for model instances

lrennels:


lrennels:
@davidanthoff @jmwingenroth  can someone approve this?  It's a follow up to the big set and update parameters update we did, just putting up some more safety barriers for people using this function (which is a speedup backdoor that only Frank and Tony use right now anyways)

src/core/connections.jl
```diff
@@ -821,7 +821,18 @@ not dirty the model, and should  be used carefully and specifically for things l
...
 """
 function update_param!(mi::ModelInstance, comp_name::Symbol, param_name::Symbol, value)
-    param = mi.md.model_params[get_model_param_name(mi.md, comp_name, param_name)]
+
+    model_param_name = get_model_param_name(mi.md, comp_name, param_name)
+    param = model_param(mi.md, model_param_name)
+
+    is_shared(param) && error("$comp_name:$param_name is connected to a ",
+            "a shared model parameter with name $model_param_name in the model, ",
+            "to update the shared model parameter please call `update_param!(mi, $model_param_name, value)` ",
```
jmwingenroth, line 830:
```suggestion
    is_shared(param) && error("$comp_name:$param_name is connected to",
            "a shared model parameter with name $model_param_name in the model.",
            "To update the shared model parameter, please call `update_param!(mi, $model_param_name, value)` ", 
```

jmwingenroth (approved):

PR #843, opened 2021-09-22
Fix bug in type checking

lrennels:


davidanthoff (approved):
Not THE most elegant fix ever, but I'm sure it will do for now :)

lrennels:
Yea I really don't like it, we need to rethink a lot of the re-written type checking that we do it's far too involved but I need this for now.

Issue #855, opened 2021-10-15
The d parameter is not typed enough


davidanthoff:
It seems that the `d` parameter to our `run_timestep` function (and presumably also the `init` function) is not properly typed, leading to type instabilities down the road.

lrennels:
We can probably use `NamedTuple` instead of `OrderedDict` for the dimensions dictionary, and perhaps other spots in the `Mimi` code like components, variables, and parameters dictionaries in order to get some more performance and from type stability inherent to those types.

PR #876, opened 2021-11-15
Use a named tuple to pass dimension dict to functions

lrennels:


lrennels:
starts to address #855

src/core/defcomp.jl
```diff
@@ -24,7 +24,7 @@ function _generate_run_func(comp_name, module_name, args, body)
     func = :(
         global function $(func_name)($(p)::Mimi.ComponentInstanceParameters, 
                                      $(v)::Mimi.ComponentInstanceVariables, 
-                                     $(d)::Mimi.DimValueDict, 
+                                     $(d)::Mimi.DimValueNamedTuple,
```
davidanthoff, line 27: I would just not type these arguments at all here, you can just leave this as `$(d)`. The type of this is not required for dispatch, so really it has no purpose here. That is true for all of the functions below.

src/core/instances.jl
```diff
@@ -351,14 +351,14 @@ function Base.run(mi::ModelInstance, ntimesteps::Int=typemax(Int),
 
     clock = Clock(time_keys)
 
-    # Get the dimensions dictionary
-    dim_val_dict = DimValueDict(dim_dict(mi.md), clock)
+    # Get the dimensions named tuple
+    dim_val_named_tuple = DimValueNamedTuple(dim_dict(mi.md), clock)
```
davidanthoff, line 355: I would just create a named tuple directly here, without any wrapper type at all.

src/core/types/instances.jl
```diff
@@ -49,22 +49,22 @@ end
 
 # A container class that wraps the dimension dictionary when passed to run_timestep()
 # and init(), so we can safely implement Base.getproperty(), allowing `d.regions` etc.
-# All values in the dictionary are vectors of Ints, except the `:time` value, which is a
+# All values in the named tuple are vectors of Ints, except the `:time` value, which is a
 # vector of AbstractTimesteps, so that `d.time` returns values that can be used for indexing
 # into timestep arrays.
-struct DimValueDict <: MimiStruct
-    dict::Dict{Symbol, Union{Vector{Int}, Vector{AbstractTimestep}}}
+struct DimValueNamedTuple <: MimiStruct
```
davidanthoff, line 55: So this won't solve the problem, because the new type is still not parameterized properly. One way to do this would be to write
```
struct DimValueNamedTuple{T<:NamedTuple} <: MimiStruct
  nt::T
end
```
But I think we can just remove this type entirely and simply pass a named tuple directly to all the user provided functions.

davidanthoff (changes requested):

src/core/instances.jl
```diff
@@ -351,14 +351,20 @@ function Base.run(mi::ModelInstance, ntimesteps::Int=typemax(Int),
...
 
-    # Get the dimensions dictionary
-    dim_val_dict = DimValueDict(dim_dict(mi.md), clock)
+    # Get the dimensions Named Tuple from the dimension dictionary which will be 
+    # passed to run_timestep() and init(), so we can safely implement Base.getproperty(),
+    # allowing `d.regions` etc.
+    # All values in the named tuple are vectors of Ints, except the `:time` value, which is a
+    # vector of AbstractTimesteps, so that `d.time` returns values that can be used for indexing
+    # into timestep arrays.
+    d = Dict([name => (name == :time ? timesteps(clock) : collect(values(dim))) for (name, dim) in dim_dict])
```
lrennels, line 360: @davidanthoff this is obviously not necessary, going from a `Dict` to a `NamedTuple`, but I was having a little trouble getting list comprehension to work properly going straight to a `NamedTuple`, any fancy ideas? I can play with it or just allocate and then iteratively append or push.

lrennels: Nevermind got it

src/core/instances.jl
```diff
@@ -351,14 +351,19 @@ function Base.run(mi::ModelInstance, ntimesteps::Int=typemax(Int),
...
-    # Get the dimensions dictionary
-    dim_val_dict = DimValueDict(dim_dict(mi.md), clock)
-
+    # Get the dimensions Named Tuple from the dimension dictionary which will be 
+    # passed to run_timestep() and init(), so we can safely implement Base.getproperty(),
+    # allowing `d.regions` etc.
+    # All values in the named tuple are vectors of Ints, except the `:time` value, which is a
+    # vector of AbstractTimesteps, so that `d.time` returns values that can be used for indexing
+    # into timestep arrays.
+    dim_val_named_tuple = NamedTuple([name => (name == :time ? timesteps(clock) : collect(values(dim))) for (name, dim) in dim_dict(mi.md)])
```
davidanthoff, line 360: Just a small little thing: you can avoid the allocation from list comprehension by using a generator. Won't make any difference, I'm sure, but still. It would then be written as

```
NamedTuple(name => (name == :time ? timesteps(clock) : collect(values(dim))) for (name, dim) in dim_dict(mi.md))
```

So you just remove the `[]` that differentiates a generator from a list comprehension, and that leads to the allocation of an extra temporary array when you use `[]`.

davidanthoff (approved):
This all looks correct to me. Maybe we should do one performance test with and without this PR, and if that is OK you can merge?

lrennels:
Regression testing currently, the results for n = 100 without SLR and Mimi base is ~50 seconds, the same runs for this branch is ~45 seconds so we're good to go and I'll merge.

PR #887, opened 2021-11-19
Add payload2 feature

davidanthoff:


lrennels (approved):
@davidanthoff looks good to me, I'd probably vote at some point for a name like `payload_shallow` or something ... but I'd have to think about what would be a good name :).  This is very internal for now anyways so not too worried about it.

src/mcs/mcs_types.jl
```diff
@@ -183,15 +185,25 @@ mutable struct SimulationInstance{T}
...
         self = new()
         self.trials = 0
         self.current_trial = 0
         self.current_data = nothing
+
+        # This is a bit convoluted, but we want to make sure to exclude payload2 from the deepcopy
+        # call here
+        payload2 = sim_def.payload2
+        sim_def.payload2 = nothing
         self.sim_def = deepcopy(sim_def)
```
lrennels, line 201: Just so I'm clear here, what we're doing is (1) deep copy the `sim_def` without payloads  (2) add the `payload` and the `payload2` (3) replace the `payload` in the `sim_def`  with a deep copied version

Issue #898, opened 2022-03-31
Resizing Dimension --> Parameter remains same size and retains old values


lrennels:
```
@defcomp mycomp begin
    mypar = Parameter(index=[country])
    function run_timestep(p,v,d,t)
        if is_first(t)
            println(p.mypar)
        end
    end
end

m = Model()
set_dimension!(m, :time, 2000:2005)
set_dimension!(m, :country, ["USA", "MDV", "ALB"])
add_comp!(m, mycomp)
update_param!(m, :mycomp, :mypar, [1,2,3])
run(m)
```
julia> run(m)
Union{Missing, Float64}[1.0, 2.0, 3.0]

```
set_dimension!(m, :country, ["USA", "MDV"])
update_param!(m, :mycomp, :mypar, [100,200])
run(m)
```
julia> run(m)
Union{Missing, Float64}[100.0, 200.0, 3.0]

lrennels:
A question here is what is the _desired_ behavior?  Resetting the time dimension is possible as long as the new time period lies within the original time period.  This will of course `dirty` the model, parameters with a `time` dimension need to be reset, and then the model rerun to get new values.  You can still access `m[:component, :parameter]` from the `ModelInstance`, and a parameter with a `time` dimension will have missing padded in the new times, or be clipped to the old times, but the data will be from the previous run.

Now taking a case like redefining the `country` dimension from `["USA", "MDV", "ALB"] to `["USA", "MDV"]`.  We have no promise that the new set of strings will be a subset of the old, so we cannot as easily use matching to clean up the existing parameters and variables. One could say `set_dimension!(:country, ["X", "Y", "Z"])`, rerun the model, and results would be the same.

Should we allow redefinition of dimensions other than time? We likely want to for flexibility. Should we add a warning about a dirty model with redefined dimensions?

PR #905, opened 2022-06-15
Reset parameters when dimension is changed

lrennels:
This PR handles #898


lrennels:
A question here is what is the _desired_ behavior? 

When we reset `:time`, our solution is to map the original set parameter values to the array of new ones, and pad the rest with `missing`. The model is set to `dirty`, and upon rerun will pick up this new parameter Array. The model instance will of course hold old values until rerun.

Now taking a case like redefining the `country` dimension from `["USA", "MDV", "ALB"]` to `["USA", "MDV"]`.  We have no promise that the new set of strings will be a subset of the old, so we cannot as easily use matching to clean up the existing parameters. One could say `set_dimension!(:country, ["X", "Y", "Z"])` ... length hasn't changed but labels have.

Solutions could be: 
1. parameters with that dimension are completely reset to `nothing` parameters and the user must update them again
2. _if_ the size of the dimension has changed, the array parameters with that dimension are reset to all `missing`
3. something else

Right now I have implemented (2)

src/core/dimensions.jl
```diff
@@ -105,8 +106,16 @@ function set_dimension!(ccd::AbstractCompositeComponentDef, name::Symbol, keys::
         redefined && (ccd isa ModelDef) && _pad_parameters!(ccd)
     end
 
-    return set_dimension!(ccd, name, dim)
+    newdim = set_dimension!(ccd, name, dim)
 
+     # check if (1) redefining dim (2) modifying a ModelDef (3) length of dimension has changed
+     if name!== :time && redefined && (ccd isa ModelDef) && dim_length_change
+        @warn("The new $name dimension keys have a different length than the original, parameters using that dimension are reset to `missings`, user will need to call `update_param!` to set them.")
```
lrennels, line 113: Will remove the warning, useful for testing.

lrennels:
Update: we want this option: **parameters with that dimension are completely reset to nothing parameters and the user must update them again**

PR #957, opened 2023-05-14
Adds MarginalInstance struct

jrising:
Currently, one can only run `build!` on a `MarginalModel`, not `build`. The second is useful for Monte Carlo runs because it gives a `ModelInstance` where parameters can be updated without requiring new building. This extends that feature to `MarginalModel`s.

A `MarginalInstance` is a pair of `ModelInstance`s, with the delta so that marginal values can be retrieved. It also has a (currently minimal) set of associated functions so that Monte Carlo code that currently takes a `Model` can just be generalized to `Union{Model, MarginalModel}` and the associated code still works.

In the diffs below, sorry that my editor removed spaces at the end of lines. If it's a problem, I can fix that.

An associated PR to MimiFAIRv2 that uses this change will come next.

### DIFF
--- a/src/core/build.jl
+++ b/src/core/build.jl
@@ -1,12 +1,12 @@
 """
     _substitute_views!(vals::Array{T, N}, comp_def) where {T, N}
 
-For each value in `vals`, if the value is a `TimestepArray` swap in a new 
-TimestepArray with the same type parameterization but with its `data` field 
+For each value in `vals`, if the value is a `TimestepArray` swap in a new
+TimestepArray with the same type parameterization but with its `data` field
 holding a view of the original value's `data` defined by the first and last
 indices of Component `comp_def`.
 """
-# helper function to substitute views for data 
+# helper function to substitute views for data
 function _substitute_views!(vals::Array{T, N}, comp_def) where {T, N}
     times = [keys(comp_def.dim_dict[:time])...]
     first_idx = findfirst(times .== comp_def.first)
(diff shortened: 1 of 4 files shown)


jrising:
Here's the MimiFAIRv2 PR: https://github.com/FrankErrickson/MimiFAIRv2.jl/pull/16

lrennels:
@jrising this seems reasonable to me, am I right that in the diff most of the changes are the line spacing removal and the main functional change is just the addition of the method?

@davidanthoff does this look alright to you?

jrising:
@lrennels I only checked in files with changes, so there are minor additions to each of those files. Here's the list of actual additions:
 - src/core/build.jl
function build(mm::MarginalModel)
function Base.run(mi::MarginalInstance; ntimesteps::Int=typemax(Int))

 - src/core/connections.jl
function update_param!(mi::MarginalInstance, comp::Symbol, param::Symbol, value)
function update_param!(mi::MarginalInstance, param::Symbol, value)

 - src/core/model.jl
function dim_count(mi::MarginalInstance, dim::Symbol)

 - src/core/types/model.jl
struct MarginalInstance
function Base.getindex(mi::MarginalInstance, comp_path::Symbol, name::Symbol)

I tried to put these where the equivalent code is for MarginalModels or ModelInstances. There are probably other functions one might want for MarginalInstance to support, but these were the ones I found necessary.

lrennels:
Sorry for the delay @jrising I'll look at this this week, I think all should be fine I just want to run our dependency tests as well.  I will note that I think I won't advertise this use for now, only because ew don't have as many safetys put onto the`Instance` structures so they would be dangerous to use if you aren't more of an expert (as of course you are :) ). 

For example I was reminded the hard way recently that if you run `update_param!` on a `ModelInstance` and the size of the array you update is not the same as the one you are updating ... it uses `copyto!` so you won't get an error that warns you that you are only replacing part of the array. Makes for some super confusing behavior :)

PR #961, opened 2023-05-20
Add find_type_instabilities

davidanthoff:
Usage is:
```julia
Mimi.find_type_instabilities(m, :compname)
```
and that will call `code_warntype` on the `run_timestep` method for that component.

In theory one can also use this with https://github.com/JuliaDebug/Cthulhu.jl by doing
```julia
Mimi.find_type_instabilities(m, :compname, Cthulhu.descend)
```
but in practice that doesn't seem to work together with our way of generating these `run_timestep` methods from our macro.

I would suggest we don't document this for now and don't consider it part of the "public" API, until we have figured out whether this is actually working well or not. We might also try to get this to work with Cthulhu properly.


lrennels (commented):
Alter creation of dim_dict to avoid nothing error.

davidanthoff:
I only tried this after running the model, maybe at that point the dims are there? But if we can get this to work without running, even better.

lrennels:
Interesting, at first I was doing 
```
m = get_model
Mimi.build!(m)
````
to get a built model but not run it because the migration model is just so slow, but I also tried it with a run `MimiGIVE` and got the same error? Did yours work for a run model?

File: src/utils/diagnostics.jl [hunk 1]
```diff
@@ -0,0 +1,18 @@
...
+
+    dim_dict_keys = collect(keys(getproperty(mi.comps_dict[comp].comp_id.module_obj, mi.comps_dict[comp].comp_id.comp_name).dim_dict))
+    dim_dict = OrderedDict([k => m.md.dim_dict[k] for k in dim_dict_keys])
+
+    f = mi.comps_dict[comp].run_timestep
+    v = typeof(mi.comps_dict[comp].variables)
+    p = typeof(mi.comps_dict[comp].parameters)    
+    d = typeof(NamedTuple(name => (name == :time ? timesteps(clock) : collect(values(dim))) for (name, dim) in dim_dict))
+
+    tool(f, (p, v, d, Int))
```
lrennels, line 17: @davidanthoff why does putting `Int` here work? I was hoping to make very similar function that returned just `run_timestep` so I could run it with @profileview` ... but seem to get errors when I try to just run:

```
f(p,v,d,Int)
```

am I missing something obvious here? Is the `Int` useful because of the specific `code_warntype` tool ... and I need to be fancier to just return the function?

lrennels: I think we want to put a `AbstractTimestep` type in there to get much more accurate results, for example
```
Mimi.FixedTimestep{1750, 1, 2300}
```
we should be able to get that from the `clock` or something, that is parameterized by the Timestep type I think, I'll take a look tomorrow.

PR #968, opened 2023-07-15
Improve performance

davidanthoff:
There are two commits here that are unrelated.

The one about the `deepcopy` I'm especially not 100% sure. But I think I followed all code paths, and I think previously the deep copy that was created originally was then actually never (except for the one case where I added a `deepcopy` returned to user, but was essentially always a temporary copy, in which case it seems unnecessary to me.

### DIFF
--- a/src/utils/getdataframe.jl
+++ b/src/utils/getdataframe.jl
@@ -24,7 +24,7 @@ function _load_dataframe(m::AbstractModel, comp_name::Symbol, item_name::Symbol,
     end
     paths = m isa MarginalModel ? _get_all_paths(m.base) : _get_all_paths(m)
     comp_path = paths[comp_name];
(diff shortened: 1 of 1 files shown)


lrennels (commented):
@davidanthoff this seems fine to me, certainly the `collect` call.  Looking through the use of `deepcopy`, my guess is that it was a safety to make sure that we don't pass around the model's data by reference, alter it in a DataFrame, and end up with a mutated model object.  I think with your change this still wouldn't happen, all the uses of `data` (other than the one you updated on line 49) mutate the array in some way that allocates a new one, but I could see it feeling fragile.

I think for now it's fine, or we could just `deepcopy` the DataFrame before returning it ... or add more calls to `deepcopy` as guard rails but I think that probably just undoes the performance fix you made :)

davidanthoff:
Yes, that `deepcopy` was the problem for performance. If you approve, I'll merge.

lrennels (approved):

PR #988, opened 2024-03-10
Fix small bug in Base.run for models and instances

jonas-hass-bonne:
The `Base.run` function allows for an optional keyword argument `dim_keys`, which (as far as I can tell) allows the user to specify a subset of time periods which the model should be run for.

This keyword argument has the following type-declaration:
`Union{Nothing, Dict{Symbol, Vector{T} where T <: DimensionKeyTypes}}`

However, this type declaration doesn't recognize a supplied value of type `Dict{Symbol, Vector{Int64}}` as valid since 
`Dict{Symbol, Vector{Int64}} <: Dict{Symbol, Vector{T} where T <: DimensionKeyTypes}` returns a value of `false`, although this is presumably the intended use of the argument.

This PR fixes this by changing the type-declaration to:
`Union{Nothing, Dict{Symbol, Vector{T}} where T <: DimensionKeyTypes}`, which ensures that arguments of the type `Dict{Symbol, Vector{Int64}}` are recognized as being valid.

This change is made in both the `src/core/model.jl` and `src/core/model.jl` files

### DIFF
--- a/src/core/instances.jl
+++ b/src/core/instances.jl
@@ -336,7 +336,7 @@ end
 Run the `ModelInstance` `mi` once with `ntimesteps` and dimension keys `dimkeys`.
 """
 function Base.run(mi::ModelInstance, ntimesteps::Int=typemax(Int),
-                  dimkeys::Union{Nothing, Dict{Symbol, Vector{T} where T <: DimensionKeyTypes}}=nothing)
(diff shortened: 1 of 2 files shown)


davidanthoff (approved):
I don't understand why the original thing doesn't work, but if this fixes it, then lets go with it! Thanks for the PR :)

PR #999, opened 2024-06-21
WIP Use Different Approach to Check for Broadcasting in Perturb Param

lrennels:
@jrising work in progress but I think this will solve https://forum.mimiframework.org/t/setting-a-distributions-over-large-dimensions-in-defsim/277/10

- [x] run Mimi tests locally
- [x] test on PAGE2020 (does fix the original error, the `init` function now errors but happy to help with that too)
- [x] test on MimiGIVE including regression testing to make sure all pass

### DIFF
--- a/src/mcs/montecarlo.jl
+++ b/src/mcs/montecarlo.jl
@@ -321,6 +321,10 @@ function _param_indices(param::ArrayModelParameter{T}, md::ModelDef, i::Int, tra
     for (dim_name, dim_values) in zip(pdims, tdims)
         dim = dimension(md, dim_name)
         dim_indices = dim[dim_values]
+
+        # reduce a singleton Vector to just an Integer
+        isa(dim_indices, Vector{Int64}) && size(dim_indices) == (1,) ? dim_indices = dim_indices[1] : nothing
+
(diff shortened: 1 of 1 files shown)


jrising:
This seems like the right solution to me. Did you figure out if it's related to String vs. String15's? I thought it was due to my use of a single dimension when you just used this for double dimension parameters.

lrennels:
No I think jt is the String15, because I use this for single dimensions too, that’s how I did my comparison I ran this step by step alongside the same for the mortality Beta in GIVE and it seems String15 returns a vector for a single index when String returns an Int.  I use CSVFiles and I have never seen that load a String type other than the base String. 

But I think this solution gets closer to the heart of when Julia says you should broadcast so that’s better regardless? Happy to keep plugging a bit.

lrennels:
We also used to use a recursive approach but I found that a bit convoluted.

lrennels:
Idea: can se use `Base.setindex_shape_check` here?

lrennels:
@davidanthoff it would be good to look through this together, I'm not sure I like the current approach but it should work and passes the tests.

lrennels:
@jrising merging, will return to this to think about a more Julia-esque arrays way to handle it

Issue #948, opened 2023-03-02
Add access to time information in the `init` function; Add Dimensions keys to `d`


lrennels:
The `init` function currently only takes `function init(p,v,d)` but does not have the time dimension, which could provide useful information to things like the `MimiRFFSPs` component which needs `start_year` and `end_year` to run properly, and these are currently `Parameters` that will not automatically update with `set_dimension!` applied to a model.

In addition, make sure that the dimensions or `d` object includes both the keys and values (names and integers) for indices to give richer information to the component functions.

lrennels:
https://github.com/rffscghg/MimiRFFSPs.jl/issues/31 for an example of where this could be useful

lrennels:
bumping this -- would be useful for multiple teams -- I'll prioritize

@jrising 
@djsmithumn

lrennels:
Getting started with this issue, we see in `instances.jl`
```
# Get the dimensions Named Tuple from the dimension dictionary which will be 
    # passed to run_timestep() and init(), so we can safely implement Base.getproperty(),
    # allowing `d.regions` etc.
    # All values in the named tuple are vectors of Ints, except the `:time` value, which is a
    # vector of AbstractTimesteps, so that `d.time` returns values that can be used for indexing
    # into timestep arrays.
    dim_val_named_tuple = NamedTuple(name => (name == :time ? timesteps(clock) : collect(values(dim))) for (name, dim) in dim_dict(mi.md))
    
    # recursively initializes all components
    init(mi, dim_val_named_tuple)

    while ! finished(clock)
        run_timestep(mi, clock, dim_val_named_tuple)
        advance(clock)
    end
```
This is where we can start passing more information to the `dim_val_named_tuple`, remembering to change the `Type` as needed. I think the steps will be

(1) Pass the information through so that at least it can be used for lookups in things like `init` functions
(2) Once that works, add ability to Index into `Parameters` and `Variables` with these keys ie. `p.myparam[:tau, :USA]`

lrennels:
Another option is to fill in the keys for the component dimension upon `build!` of the component instance, or wherever the above initialization happens, so that the Component Instance knows it's own keys and then pass these to a built-in function like `get_dim_keys` ...

lrennels:
Another use case example: https://forum.mimiframework.org/t/setting-subset-for-regions/307/4

PR #1017, opened 2024-10-23
WIP Add Dimension Key Access to `run_timestep` function

lrennels:
Handling part of #948 (also mentions access to `t` in `init` but save that for a second PR)

The goal of this PR is to add access to the dimension **keys** from within components. Currently for a `run_timestep(p,v,d,t)` function, `d.dim_name` will return a vector of the `Type` needed to **index** into Arrays for the dimension, namely `Int`s for all dimensions except `time` for which it is our custom `Timestep` type.

While this is useful for a loop like
```Julia
for d in d.countries
    v.var1[d] = p.par1 * p.mulitplier[d]
end
```
there are times a user would like to check the actual keys, I think the core example being something like 
```Julia
for d in d.countries
    if d == "USA" # FAILS
        # do something
    else
        # do something else
    end
end
```
Note the line above fails because `d` is an Integer (eg. in GIVE we know it is 174, the index of the key "USA" in the countries dimension vector).

One way to pass information to `run_timestep` without breaking compatibility is to have keyword args we set during initialization, and then are in scope for the function, that the user never needs to see or manipulate. We have access to the entire `ModelInstance` at the time of running `run_timestep`.

Some decisions/TODOs

1. What should we pass? On my first try I pass a `get_dim_keys(dim_name::Symbol)` function, but we could really pass any structure that is useful. Instead of a function we could have an object called `d_keys` that is exactly the same as `d` but returns the keys?
2. Add this to `init` as well (?)
3. Think carefully about the time dimension keys here -- there can be offsets etc.
4. Add tests

### DIFF
--- a/src/core/defcomp.jl
+++ b/src/core/defcomp.jl
@@ -25,7 +25,8 @@ function _generate_run_func(comp_name, module_name, args, body)
         global function $(func_name)($(p), #::Mimi.ComponentInstanceParameters,
                                      $(v), #::Mimi.ComponentInstanceVariables
                                      $(d), #::NamedTuple
-                                     $(t)) #::T <: Mimi.AbstractTimestep
+                                     $(t); #::T <: Mimi.AbstractTimestep
+                                     dim_keys::Function) #::Function
             $(body...)
             return nothing
         end
@@ -47,7 +48,8 @@ function _generate_init_func(comp_name, module_name, args, body)
     func = :(
         global function $(func_name)($(p), #::Mimi.ComponentInstanceParameters
                                      $(v), #::Mimi.ComponentInstanceVariables
-                                     $(d)) #::NamedTuple
+                                     $(d); #::NamedTuple
+                                     dim_keys::Function) #::Function
             $(body...)
             return nothing
         end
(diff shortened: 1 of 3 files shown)


lrennels:
@jrising @ djsmithumn @bryanparthum I think all three of you have mentioned this to me, I'm taking an incremental approach here, although @jrising I admit at this point now that I understand the structure I could pretty easily just pass the entire `mi` -- keeping it light for now

I'm just playing with it for now

**Question:** What would your preferred way of accessing dimension information in `run_timestep` be? Here's a common use case I've seen:

```
for d in d.country
         if d == "USA" # this fails because d is an Integer
                  blah
         else
                  blip
         end
end
```
so now it could be something like ...
```
for d in d.country
         if dim_keys(:country)[d] == "USA" # this fails because d is an Integer
                  blah
         else
                  blip
         end
end
```
orrrr
for d in d.country
         if d_keys.country)[d] == "USA" # this fails because d is an Integer
                  blah
         else
                  blip
         end
end
```

**Comment by lrennels** [comment]
Current implementation (not totally happy with it though ...) would let you do ...

```Julia
for d in d.countries
    if get_dim_keys(:countries)[d] == "USA"
        # do something
    else
       # do something else
    end
end
```
we could also have something like 
```Julia
for d in d.countries
    if d_keys.countries[d] == "USA"
        # do something
    else
       # do something else
    end
end
```
or maybe something fancier ...

jrising:
The `get_dim_keys` pseudo-function looks good to me. I guess I would just suggest that in the unit test, you spin up two different models with the same component but different dimension values, to sanity-check that there are no global dependencies in the `get_dim_keys` function.

PR #1018, opened 2024-11-20
Pass simulation data via CSV file to explorer

davidanthoff:
This might work better with large simulation data than the current approach. I only implemented it for one specific plot type. If this works we might also use this approach for other figures.

Also updated vega and vega-lite to their latest versions.

### DIFF
--- a/src/explorer/assets/main.html
+++ b/src/explorer/assets/main.html
@@ -3,13 +3,11 @@
 <html>
 <header>
     
-    <script src="https://cdn.jsdelivr.net/npm/vega@3"></script>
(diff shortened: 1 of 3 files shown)


lrennels:
@davidanthoff from a code perspective this looks good to me, and all tests are passing. I also ran the `test_explorer_sim.jl` test file locally and looked at outputs to make sure things looked ok and they do as far as I can tell.

In terms of improvement to performance we should test with a model that has a fine grained spatial dimension, like countries (GIVE) or even coastal segments (CIAM). I think @corakingdon has an example case, but I can give it a try too.

PR #1020, opened 2024-11-28
Make `get_param_value` and `get_var_value` work on models

rturquier:
This should make it easier to inspect a model.

### DIFF
--- a/src/core/instances.jl
+++ b/src/core/instances.jl
@@ -91,6 +91,17 @@ function get_param_value(ci::AbstractComponentInstance, name::Symbol)
 end
 
 """
+    get_param_value(model::Model, comp_name::Symbol, param_name::Symbol)
+
+Return the value of parameter `param_name` in component `comp_name` of model `model`.
+"""
+function get_param_value(model::Model, comp_name::Symbol, param_name::Symbol)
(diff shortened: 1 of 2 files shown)


lrennels:
Hi @rturquier thanks so much for contributing! Is the goal here a functionality different than using something like the following?
```
m[:comp name, :paramname]
```
or 
```
getdataframe(m, :compname, paramname)
```
My hypothesis would be you're wanting to access these from a model that has not been run yet?

rturquier:
Thank you for your reply @lrennels! What happened is that I found `get_param_value` in the documentation, and didn't know about the `getindex` syntax. I now see that it's in one of the how-to guides, which is great. So, I guess the way forward would be to just remove `get_param_value` and `get_var_value` from the exports and documentation?  

(I don't think it's a big issue that it doesn't work before the model is run, and it would probably be best to just make it work instead of having a different method.)

lrennels:
Hi @rturquier great, yes I'm sorry those indexing methods weren't more immediately obvious, if there is somewhere in the documentation you think we can improve that do let me know. I think the primary functions users use to explore their results are
```
getdataframe
getindex
explore
```
Let us know if you have any further questions, I'll close this issue otherwise!

PR #1022, opened 2024-11-28
WIP: ExceFiles Package Test Failure -- Failed to Precompile

lrennels:
Tests seem to be failing due to the `ExcelFiles` package, which we've had bugs with in the past. It is included in only one test, so here we will remove it from the testing environment and if that successfully solves the testing errors we can try to replace the tests with `CSVFiles.jl`


lrennels:
@davidanthoff I think `ExcelFiles` is a package in `Querverse`, and seems to be causing a new precompilation failure on Mimi `main`. I removed the dependency (and the one on `ExcelReaders`) in this PR to test that those were the problem. Maybe this will settle itself out, I can try to re-run the Mimi tests in a week or so, but otherwise maybe we can look into the recompilation failure?

The test isn't really crucial to core Mimi, but I'd prefer not to remove it if possible.

davidanthoff:
Yeah, both of those packages have been broken for years. They wrap Python packages that no longer work properly. So this here looks good to me.

lrennels:
@davidanthoff okeedokee, I'll see if there's another way to run this test, or a modified version of it, using `CSVFiles`. I think it's testing our `EmpiricalDistribution` so that should be doable. Then I'll remove those two dependencies permanently from `Mimi.jl`.

davidanthoff:
You could also just save the file as a xlsx format and then use XLSX.jl to read it, which should work. xlsx has been the default format for Excel I think for 20 years now :)

lrennels:
:) true, I think I remember we originally didn't do that for the `MimiIWG` tests because switching to `.xlsx` changes the numbers a tiny bed when resaving -- and we were testing a perfect replication of the Roe Baker Distribution -- but for this it's fine if we don't

PR #1029, opened 2025-02-21
Add code_warntype

davidanthoff:


lrennels:
@jrising you may like this -- we added a quick way to check for type instability in the `run_timestep` function of any component

PR #1059, opened 2026-03-27
Remove dep on Classes and simplify type hierarchy

davidanthoff:
This removes the dependency on Classes.jl and simplifies the type hierarchy. Claude thinks this is all easy and non-breaking, but we should try this branch with a lot of different models to make sure nothing breaks.


davidanthoff:
I've tested this with MimiGIVE now and it all works, so I'm going to merge this.

PR #1081, opened 2026-09-18
Add multi-threaded Monte Carlo runs via an ntasks kwarg

davidanthoff:
Adds multi-threaded Monte Carlo runs, opted into with a new `ntasks` keyword argument on `run`. Opened by Claude (Claude Code).

```julia
si = run(sd, m, 100_000; ntasks = 8)     # or ntasks = :auto for one task per thread
```

`ntasks` defaults to 1, which is the previous serial path, unchanged.

## How it works

Rather than spawning a task per trial (a million `Task`s and their stacks, for a 1M-trial run) or splitting the trials into fixed blocks up front (imbalanced when trials take different amounts of time), `run` spawns exactly `ntasks` worker tasks. Each owns one set of model instances for its whole lifetime and claims the next unclaimed trial from a shared atomic counter. Work is balanced at the granularity of a single trial, and the number of live model instances — and so the memory the run needs — is capped at `ntasks`, which is why the number is left to the caller rather than always using every thread. Worker 1 uses the caller's own models, so only `ntasks - 1` copies are made and `si.models` still refers to models that really ran.

Extracting a trial's results is separated from storing them. Workers extract into `DataFrame`s (this is where `getdataframe`/`_get_all_paths` cost lives, so it parallelizes), and a single writer task appends them to the simulation results and drives the CSV save streams, neither of which is safe to touch from more than one task. The writer holds results that arrive early in a small reorder buffer until their turn comes, so results are stored in trial order however the workers interleave. That matters beyond tidiness: `analyze` for Sobol and Delta matches model output against the sample matrix positionally, so unordered results would silently produce wrong sensitivity indices, and streamed CSVs cannot be sorted after the fact.

Trial callbacks keep their signatures. Each worker gets a lightweight `SimulationInstance` view whose `models` are its own, while `sim_def`, the payloads and the results storage stay shared, so `sim_inst.models[1]` inside a `post_trial_func` keeps working.

## Please look at this one thing

`_copy_sim_params` did not actually snapshot array parameters. `Base.copy` of an `ArrayModelParameter` wraps a new parameter around the *same* values, so `_restore_sim_params!` was assigning an array to itself, and `+=`/`*=` perturbations of **array** parameters compounded from trial to trial rather than being applied to the original values each time. Trial `k` was getting `base * u_1 * u_2 * ... * u_k` instead of `base * u_k`. Scalar parameters were always restored correctly.

This surfaced because it is the one thing that cannot be reproduced in parallel: a parallel run gave the *correct* answer for some trials and the compounded one for others, depending on which worker picked up which trial. The fix is in this PR (`_snapshot_param`), with a regression test.

**It changes results for any simulation that perturbs an array parameter with `+=` or `*=`.** Every existing test still passes, but published numbers from such a simulation will move. Flagging it in case you would rather split it into its own PR and release.

## Verification

- All 54 test items pass, on 1, 4 and 8 threads.
- New `test/mcs/test_parallel.jl` asserts that a parallel run reproduces a serial one exactly — same values *and* same row order, in memory and byte-for-byte in the output CSVs — with no scenarios, with an `OUTER` scenario loop, and with an `INNER` one whose `scenario_func` mutates the models. It also covers a `payload` indexed by trial number, `MarginalModel`, `results_in_memory = false`, `ntasks` validation, errors from a trial reaching the caller as themselves rather than wrapped in a `TaskFailedException`, and the array-parameter restore regression.
- Results are identical for `ntasks` of 1, 2, 3, 8 and `:auto` from the same seed, and repeating a parallel run reproduces itself.
- A model with a deliberately slow `run_timestep` gets a 3.65x speedup on 4 threads.

## On random numbers

Mimi's own draws — for every sampling strategy — are all made up front by `generate_trials!` before any worker exists, so trial data is unaffected by `ntasks` and `Random.seed!` before `run` reproduces it exactly. `get_trial` now indexes the `SampleStore` by trial number instead of relying on its cursor advancing in trial order, which is what makes trials order-independent.

Model or callback code calling `rand()` itself is a different matter: those draws come from julia's task-local RNG, so each worker has its own stream and which stream a trial sees depends on which worker claimed it. Not reproducible, and different from a serial run. Per our discussion this is documented rather than worked around — no seeding machinery in this PR. The docs say so plainly, and the new test item carries a comment explaining why nothing in it may call `rand` from a component or callback.

A model holding its *own* RNG object and sharing it across trials would be a genuine data race; that is called out in the docs alongside the payload thread-safety contract.

## Also in here

- `.github/workflows/juliaci.yml` sets `JULIA_NUM_THREADS: 4` so CI actually exercises threads — without it the new test item still runs, but interleaved on one thread, which catches deadlocks and ordering bugs and not real races. This was flagged as your call in the plan; it is a one-line revert if you would rather not.
- The scenario name and output directory for each inner scenario are computed once per scenario instead of once per trial, taking a `mkpath` syscall out of the trial loop.
- Results columns for *scalar* saved data are now concretely typed rather than `Vector{Any}` (same values, cheaper over a million trials).
- `NEWS.md` has not been touched: it has not been updated since v0.8.5, so it looked deliberately retired.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


davidanthoff:
## Tested this branch against public Mimi models

I ran the branch against the public Mimi models that (a) install and work on released Mimi 1.7.1 and (b) ship their own `@defsim`, each in a throwaway environment with Mimi `dev`ed from this branch, on 4 threads. Julia 1.13, Windows, 36 cores.

For each model: build and run it on **Mimi 1.7.1** to get reference output, then on this branch check that a serial run reproduces that reference byte for byte, and that `ntasks` of 1, 2, 4, 8 and `:auto` all reproduce the serial run exactly — same values *and* same row order, in memory and in the streamed CSVs — plus a repeat parallel run, a `results_in_memory = false` run, and timing. 13 checks per model, one of which is a **negative control** that reruns from a different seed and requires the comparison to report a difference; without that a green result means nothing.

| Model | Version | Trials | Runs on 1.7.1 | Serial unchanged by PR | `ntasks` == serial | Speedup @ 4 | Peak RSS 1 → 4 tasks |
|---|---|---|---|---|---|---|---|
| MimiPAGE2009 | 5.0.0 | 5,000 | ✅ | byte-identical | ✅ | **3.30x** | 1.08 → 1.09 GiB |
| MimiPAGE2020 | 1.8.0 | 3,000 | ✅ | byte-identical | ✅ | **2.32x** | 1.22 → 1.23 GiB |
| MimiFUND | 3.13.0 | 100 | ✅ | byte-identical | ✅ | **3.35x** | 1.31 → 1.31 GiB |
| MimiGIVE | 2.2.0 | 100 | ✅ | byte-identical | ✅ | **3.21x** | 1.71 → 1.75 GiB |
| MimiFUND SCC MCS | 3.13.0 | 50 × 4 scen | ✅ | byte-identical | ❌ **see below** | 2.29x | 1.82 → 1.84 GiB |

The design holds up. No deadlocks or hangs in ~70 parallel runs, errors from a trial reach the caller unwrapped, `ntasks = :auto` works, and MCS/LHS/Sobol/Delta all give identical results serial vs parallel. Memory scales as claimed — GIVE goes 1.26 / 1.35 / 1.43 / 1.57 GiB at `ntasks` 1 / 4 / 8 / 16, about 20 MiB per extra worker.

But one real bug, now fixed on this branch.

---

### `_copy_models` dropped state held only by the built `ModelInstance`

`_copy_model` built each worker's model with `Model(m)` + `build!`, which copies the `ModelDef` and nothing else. Not everything a model carries is in its definition — code that reaches a parameter through `compinstance` writes into the instance alone. `MimiFUND.perturb_marginal_emissions!` is exactly that:

```julia
ci = compinstance(m, comp_name)
emissions = Mimi.get_param_value(ci, :add)
...
emissions[:] = new_em
```

and MimiFUND's own `run_fund_scc_mcs` calls it from its `scenario_func`. So workers 2..N ran the "marginal" model **with no emissions pulse**, their marginal damages came out exactly zero, and so did their SCC. Measured over 40 trials at `ntasks = 4`:

```
  trial   serial          ntasks=4
      1         22.707            0.0
      2          9.727          9.727
      3         26.721            0.0
      ...
  trials whose SCC came out exactly 0.0 under ntasks=4 : 30 of 40
  trials whose SCC came out exactly 0.0 serially       :  0 of 40
  mean SCC serial   = 12.734435172300701
  mean SCC ntasks=4 =  3.1528785896272074
```

No error, no warning — a 75% understatement of the SCC. The ~1/4 of trials that were right are the ones worker 1 happened to claim, since it uses the caller's own models, so *which* trials are wrong changed from run to run (a repeat gave a mean of 3.213).

Isolated without any parallelism at all:

```
  pulse on the ORIGINAL modified model: max = 0.2727272727272727
  pulse on Model(m) + build!  copy    : max = 0.0      <= what a worker task got
  marginal loss: original vs rebuilt copy: DIFFERS in 15696/16816 entries
      sum of marginal loss, original = 1372.2717819318095
      sum of marginal loss, rebuilt  = 0.0
  marginal loss: original vs deepcopy copy: IDENTICAL
```

**Fix, pushed here:** copy the model whole rather than rebuilding it.

```julia
function _copy_model(m::Model, memo::IdDict)
    return get!(memo, m) do
        m_copy = deepcopy(m)
        is_built(m_copy) || build!(m_copy)
        m_copy
    end
end
```

`is_built(m) = !(dirty(m.md) || modelinstance(m) === nothing)` already accounts for the dirty flag, so the case the old code was written for — a `scenario_func` that changes the *definition* — still gets its rebuild. I checked that explicitly, both when the `scenario_func` calls `build!` itself and when it leaves the model dirty.

After the fix: MimiFUND-SCC gives SCC values identical to serial, 0 of 40 trials zeroed, and all five suites pass 13/13 with unchanged speedups. Cost on MimiGIVE for the three copies an `ntasks = 4` run makes: `Model(m)+build!` 0.23 s / +56 MiB live heap, `deepcopy(m)` 1.22 s / +55 MiB. One extra second per run, same memory.

Also added a regression test (`test/mcs/test_parallel.jl`) that asserts `_copy_models` preserves an instance-only parameter — directly, rather than relying on the scheduler handing a trial to a worker other than the first — alongside an end-to-end serial-vs-parallel check. It fails cleanly on the previous code.

---

### Two documentation additions, both from things that actually happened

**Writing to a file from a callback.** MimiFUND's `scc_calculation` appends its result with `open(f, "a")` from inside `post_trial_func`. Of 200 expected rows: serial wrote 200, none malformed; `ntasks` 2 / 4 / `:auto` wrote 198–199, and one run left a truncated line containing just `7`. Dropped and torn writes. The thread-safety section now says so and suggests recording into an array indexed by `trialnum` and writing after `run` returns.

**Mutating a model held in the payload.** The payload bullet said "appending to a shared vector or accumulating into a shared scalar is a data race"; it now also names *running or modifying a model reached through the payload*. The motivating case is MimiGIVE's `compute_scc`, whose `post_trial_func` unpacks `ciam_base`, `ciam_modified` and CSV `streams` from `Mimi.payload2` and mutates all three — those are single shared `Mimi.Model`s and shared write streams. I could not exercise that path without patching MimiGIVE, which was out of scope, but by inspection anyone who reaches for `ntasks` there gets a genuine race. (GIVE's plain `run_mcs`, which is the configuration in the table above, has no callbacks and is clean. GIVE's marginal model is *not* affected by the `_copy_models` bug — `add_marginal_emissions!` goes through the ModelDef.)

Added one sentence to "Running trials in parallel" about the single writer task being the serial fraction, since that is visibly what caps the speedup: PAGE2020 saves 19 variables and gets 2.32x, PAGE2009 saves 12 and gets 3.30x.

---

### On the `_snapshot_param` change flagged in the description

It's a real pre-existing bug and the fix is right. On Mimi 1.7.1, with `p *= Uniform(0.8, 1.2)` on an array parameter, the effective multiplier applied to the original values over 12 trials:

```
1.7.1    [1.01915, 1.1366, 1.36335, 1.19016, 1.33941, 1.34394, 1.54001, ...]   drifting out of support
branch   [1.01915, 1.11525, 1.19949, 0.87297, 1.1254,  1.00338, 1.14589, ...]  all within [0.8, 1.2]
```

**It moves no public model's numbers**, so I don't think it needs splitting out. None of the five simulation definitions uses `+=` or `*=` on an array parameter, and the check comparing branch-serial output against Mimi 1.7.1 serial output was byte-identical for all five. It does deserve a line in the release notes.

One incidental benefit of the concrete-typed result columns: `analyze` for Sobol now works on scalar saved data. On 1.7.1 it throws `MethodError: no method matching analyze(::SobolSimulationInstance, ::Vector{Any})`; on this branch it returns, and gives identical indices serial vs `ntasks = 4`. (`analyze` draws its own bootstrap CIs, so it has to be seeded before each call or two calls on identical input differ — that tripped me up before I noticed, and is worth knowing for anyone verifying determinism.)

---

### Two things found along the way that are not this PR's

- A pre-existing **serial** bug: a `scenario_func` that changes the ModelDef without calling `build!` makes the serial path drop the first trial's parameter perturbation for each scenario. Reproduced on released 1.7.1. The parallel path gets it right, so this is the one case where parallel ≠ serial and parallel is correct. Filed separately.
- **MimiIWG** was in the original test set and had to be dropped: 1.2.0 pins `MimiFUND = "=3.8.6"` against its own `Distributions = "0.23, 0.24, 0.25"`, which is unsatisfiable on a current ecosystem, and Pkg quietly resolves down to MimiIWG 0.1.0. Nothing to do with this PR; filed on that repo. Its feature coverage (MarginalModel + `OUTER` scenario loop + payload + `post_trial_func`) was picked up by MimiFUND's SCC simulation instead — which is where the bug turned out to be.

**MimiMETA** was evaluated and excluded: it isn't a Julia package and doesn't use Mimi's Monte Carlo runner at all — `src/montecarlo.jl` builds its own draws with `Mimi.lhs` and loops over them itself, so it exercises none of this.

